PR #4, opened 2019-06-05
Allow a custom font path to be given

rhurstdialpad:
Moved from https://github.com/dialpad/design/pull/38

### DIFF
--- a/lib/scss/classes/utilities/typography.scss
+++ b/lib/scss/classes/utilities/typography.scss
@@ -23,9 +23,9 @@
         font-style: "normal";
         font-weight: $key;
         font-family: "MarkWeb";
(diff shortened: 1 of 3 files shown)


rhurstdialpad:
@hynes-dialpad Are you comfortable with this change? I'd like to get it merged so that I can get classes working in fire spotter. Currently only @ extends works.

hynes-dialpad (approved):
Looks good to me. We might want to eventually pull items like $font-path into a "things you might potentially change with every project" token file. Depends on how many of these tokens we come across.

rhurstdialpad:
@hynes-dialpad That makes sense to me. I'll create an issue for it.

PR #6, opened 2019-06-10
SCSS ---> LESS

hynes-dialpad:
This converts Dialtone from using a SCSS CSS pre-processor to a Less CSS pre-processor.

There are a few reasons for this:

 1. Less doesn't require tokens (or variables as they call them) to be declared in linear order. This makes importing and breaking up files a lot more forgiving with Less. As long as a mixin, variable, or variable map is referenced somewhere, it'll compile.
 2. Less provides better nesting controls.
 3. Less allows any class to be used within another class, copying the first class's properties (vs. extending them, which the only way you can do it with SCSS other than making it a mixin). This allows you to declare the class `.flex { display: flex; }`, for example; and then include it in `.another-class { .flex; }`. This would output: `.flex { display: flex; } .another-class { display: flex; }`.
 4. Less, as of v3.5, allows for variables maps which can be combined with the `each()`function like in SCSS.

As a part of this conversion, the following items were changed:

 - All silent definitions were removed. Less doesn't really have these (or need them) since you can insert any class within another class to copy its properties.
 - The tokens file was renamed to variables since this is the way Less refers to them.
 - All variables are still available within a `dialtone-variables.less` file, though I'm unsure why you would want to include just the variables (and mixins) without the classes.
 - All responsive classes have been removed for now. We don't use them yet. I still need to create a mixin to help generate those classes. For the sake of time, I pushed that off.


rhurstdialpad (approved):
Looks good to me 👍🏻

rhurstdialpad:
Very cool 👍

PR #10, opened 2019-07-12
swap BEM separators: B--E__M => B__E--M

superfreddialpad:
I was about to jump into converting UC to use the less-based version of dialtone, when something I missed before suddenly jumped out at me: while I had previously noticed the syntax appearing in newer styles (which is what got to start looking into it and eventually proposing we make it official) only now did I realize that you've got the separators backwards:

```lang=less
// icon is an element of the d-btn block
.d-btn--icon { ... } // swapped
.d-btn__icon { ... } // correct

// striped is a modifier of the d-table element
.d-table__striped // swapped
.d-table--striped // correct
```

I don't think it's feasible or even necessarily desirable for us to dive headlong into the whole BEM philosophy (one of my favorite parts about it is that we get a lot of value even if we never do) but I do think that we should be adhering to the syntax. I'm sure the diversion was accidental, but I think it's important to correct it while we can.

Fortunately, there's actually not that much to change; I started to write a `sed` script, but then I realized that it was trivial to narrow down the required changes to like ~70 lines, so I just did it manually. I would ask that you give this a once over to make sure that I didn't snag any false positives, and merge it if it does.

I'm also going to add a few comments that I hope can offer insight into what I hope to achieve with a more purposeful/explicit approach to class names.

tagging @rhurstdialpad as well, since this will ostensibly effect Dialtone in Dialpad, and he may have to update some names on his side as well. Obviously it's important to make sure everyone's ducks are in their correct rows before we publish a new dialtone package.

I'm not used to creating pull requests/reviews in github so apologies if I screwed something up with the sending of this request.


lib/less/utilities/grid.less
```diff
@@ -52,17 +52,17 @@
 //  ============================================================================
 //  $   GRID LAYOUTS
 //  ----------------------------------------------------------------------------
-.d-gl__col {
+.d-gl--col {
```
superfreddialpad: I originally thought the hits in this file (`--col`, `--sidebar-content` and `--header-content`) were false positives, but I'm pretty sure this is correct: these actually *are* modifiers.

Which leads me to a direct example of something I've been trying to emphasize in UC: these names are not quite up to snuff. They're not super-egregious by any means, but they tripped me up when I saw them, and I had to think harder than I might have otherwise in order to resolve the confusion.

I don't think this suggestion is infallible by any means, but I think these should be called something like `.d-gl--with-header-content`  and `.d-gl--with-sidebar-content`.

Like I said, that's a subtle , but it's related to a broader topic quite a bit more important: `d-gl` should really be something more like `d-grid-layout`. Same goes for utility names like `m`, `bt`, `mb`, `brt`, `p` etc.

While I lack the popedom required to grant infallibility per se, this one might as well be written in stone: generally accepted code standards say that one and two letter variable names (which is essentially what class names are) are a major no-no.

now, even in languages where that principle is more typically applied, there are notable exceptions to that rule: `i` for counters, `e` for events or errors, etc. So it may be that some of the names I'm giving as an example here are reasonable exceptions that seem unreasonable to me because of my lack of familiarity with the form.

That said: what I really want to emphasize here is:

A) short names save you few milliseconds of typing at the cost of reading-time, both to future-you and other authors (especially those from other disciplines or newbies), which is not a worthwhile tradeoff.

B) long class names are a bummer when you need to cram a lot of them into a classList, which may be a justification for artificially shortened names (i.e. utility classes), but that need is significantly reduced with less (and to some extent, vue templates).

I know that I probably sound like a broken record there, but I honestly don't think it's possible to over-emphasize that such terse names are not only harmful, they are also obsolete. Even in the worst-case scenario, where class names are so long they're starting to affect performance, we could easily minify them before deployment.

Anyway, that comment was much longer than I thought it would be, but cuts to the core of what I'm feeling about DT at the moment. I really want to revisit this (broader) issue with the whole team at some point: it's not too late to correct course and (as always) I am more than happy to volunteer my services in automating name inflation.

hynes-dialpad: > B) long class names are a bummer when you need to cram a lot of them into a classList, which may be a justification for artificially shortened names (i.e. utility classes), but that need is significantly reduced with less (and to some extent, vue templates).

This is a good point and one I hadn't considered. Previously most people I worked with wanted terse class names, but in those circumstances we were placing them all into the HTML object's class property. So `m0 p16 d-block ba bc-navy bgc-navy-light` was preferred to something like `margin0 padding16 display-block border-all border-color-navy background-color-navy-light`. I have a better idea what's going on in the second set, though it's a lot longer. But to your point: these won't be used on their own too often (if at all), compiled instead under a semantic class.

I'm fine for going longer with the names for now.

hynes-dialpad (approved):

PR #13, opened 2019-07-26
remove global non-namespaced styles

dbecherdialpad:


lib/less/dialtone.less
```diff
@@ -27,8 +27,6 @@ body {
 }
 
 body {
-    color: @color--dark;
     font-size: @base-font-size; // [2]
     font-family: @font-family--mark;
```
bendialpad, line 31: Do we need these or can they go also? Or, do they have no impact on existing styles either way?

lib/less/dialtone.less
```diff
@@ -27,8 +27,6 @@ body {
 }
```
bendialpad, line 27: Is this one going to change anything?

bendialpad (approved):
Just 2 comments !

lib/less/dialtone.less
dbecherdialpad: no, we already set markweb on <body>. the root font size doesn't appear to cause any issues that I've seen, but if it does I think we should reset it in the app css since this is a pretty standard default.

lib/less/utilities/typography.less
```diff
@@ -14,20 +14,6 @@
 //  • Text transform
 //  • Text alignment
 //  • Line height
-//
-//  ============================================================================
-//  $   FONT FACE
-//  ----------------------------------------------------------------------------
-each(@font-family-weights, {
```
dbecherdialpad, line 21: to be clear: I think we should do this, I just want to do it as a separate change since it'll require a few changes to our app css.

lib/less/utilities/typography.less
```diff
@@ -18,16 +18,19 @@
 //  ============================================================================
 //  $   FONT FACE
 //  ----------------------------------------------------------------------------
-each(@font-family-weights, {
+// Can override @define-font-face to false to prevent font definitions below.
+@define-font-face: true;
```
bendialpad, line 26: Probably worth adding a quick comment why we're doing this.

PR #33, opened 2019-11-15
Update build script to export Vue SVG files

hynes-dialpad:
Per our discussion in today's meeting, this updates `gulpfile.js` so that it also exports the SVG files into their own Vue component files.

Each SVG is wrapped in a `<template>` tag.

Also the SVG file names are renamed. Currently the `SVG` files are lowercase, separated with single dashes. The script name changes this to PascalCasing, prepending the word "Icon" to the beginning of every `.vue` file. An example:

```
IconAdd.vue
IconAttachFile.vue
```


dbecherdialpad (approved):
Awesome! Are these all in the documentation now?

hynes-dialpad:
@dbecherdialpad Ah, no. I should add the bit about Vue now.

docs/_data/icons.yml
```diff
@@ -1,264 +1,396 @@
...
+  vue: IconArrowDropDown
 - icon: Arrow, Drop Up
   file: arrow-drop-up
+  vue: IconArrowDropUp
 - icon: Arrow, Forward (iOS Style)
   file: arrow-forward--ios
+  vue: IconArrowForwardIos
 - icon: Arrow, Forward
   file: arrow-forward
+  vue: IconArrow Forward
```
dbecherdialpad: space

lib/build/less/dialtone-globals.less
```diff
@@ -14,3 +14,41 @@ body {
...
+
+    width: 3.6rem;
+    height: 3.6rem;
+}
+.d-svg--xl {
+    .d-svg();
+
+    width: 4.8rem;
+    height: 4.8rem;
+}
```
dbecherdialpad, line 98: while you're at it maybe should think about adding `variables/transitions.less` to this globals file? It similarly needs to be globally defined.

lib/dist/css/dialtone.css
```diff
@@ -6251,3 +6248,36 @@ body {
...
   font-size: 1.6rem;
 }
+.d-svg {
+  width: 2.4rem;
+  height: 2.4rem;
+  fill: currentColor;
+}
+.d-svg--xs {
+  width: 2.4rem;
+  width: 1.4rem;
```
dbecherdialpad, line 6026: weird what's going on with width/height being defined twice?

dbecherdialpad (approved):
this is awesome, thanks for your hard work on getting it done so quickly!

docs/_data/icons.yml
hynes-dialpad: This has been fixed.

lib/dist/css/dialtone.css
hynes-dialpad: I think it is from the fact its inheriting everything from d-svg.

PR #36, opened 2019-11-18
Update toasts CSS and provide better documentation

hynes-dialpad:


dbecherdialpad (approved):
I don't think DP actually uses these toasts anywhere yet so should be fine

PR #40, opened 2019-11-27
Fix button size

hynes-dialpad:
This fixes #39.

@vincentdialpad: This should fix the issue you were having with button sizes. The button size classes were placed _before_ the various button styles, so were not applying correctly. This is what you should see in the next Dialtone version:

image


vincentdialpad:
Awesome! Thanks!

PR #44, opened 2019-12-05
Fix selected and disabled button states

hynes-dialpad:
This PR removes the various `--selected` and `--disabled` button classes previously in Dialtone. Previously the pseudo state and an alternate class was provided, but in most instances the class instance is not needed. In order to have these classes required the button less code to be overly complicated. This simplifies the button code and ensures button styles are applied correctly.

This fixes #42 and #20.


vincentdialpad:
This would be a breaking change for anywhere using the classes exclusively. I believe there was a request to allow the disabled class to be applied separately by @mraj to still trigger events.

hynes-dialpad:
@vincentdialpad yeah I saw that in the earlier requests. I just pushed a change to add those classes back in as well as loading button classes as well. I don't think we need the `.d-btn--selected` classes.

vincentdialpad (approved):
I guess an argument for requiring a base d-btn class is it would make maintaining these different states easier.

hynes-dialpad:
@vincentdialpad Yeah, it would, but we've moved this direction so specific states could be applied while applying one class. I guess we could always re-visit it in the future. Maybe when we create a Vue button component? ¯\\_(ツ)_/¯

PR #77, opened 2020-05-29
Fix navy-lighter hex value

tedgoas-dialpad:
The hex value for navy-lighter was incorrect on the docs site.


hynes-dialpad (approved):
👍 Thanks for finding this @tedgoas-dialpad

PR #78, opened 2020-06-03
Theming

hynes-dialpad:


lib/build/less/dialtone.less
```diff
@@ -13,3 +13,6 @@
 @import "dialtone-variables";
 @import "dialtone-globals";
 @import "dialtone-classes";
+
+//  --  THEMES
+@import "tmobile";
```
tedgoas-dialpad: @hynes-dialpad If someone inspected our code, could they find our client's name? If they did, would it be a privacy concern?

hynes-dialpad: Good point. It's a proof of concept, but I can scrub the name.

brianfirespotter: i'm not too worried about anyone finding it right now and even knowing what it means

PR #92, opened 2020-07-14
Add breadcrumbs

hynes-dialpad:
This adds breadcrumb visual styles into Dialtone. Each `d-breadcumb` element adds a `/` via `:before` element (except for the last one) so we don't need to add this character into the HTML.


docs/components/breadcrumbs.html
```diff
@@ -0,0 +1,61 @@
...
+                <a href="#" class="d-breadcrumb--inverted">Root</a>
+                <a href="#" class="d-breadcrumb--inverted">Section</a>
+                <a href="#" class="d-breadcrumb--inverted">Section</a>
+                <a href="#" class="d-breadcrumb--inverted">Section</a>
+                <a href="#" class="d-breadcrumb--inverted--selected">Current Page</a>
+            </div>
+        </div>
+        <div class="dialtone-example__code">
+{% highlight html linenos %}
+<div>
```
tedgoas-dialpad, line 44: 💡 Idea: What if we did something more like this?

```
<div class="d-breadcrumbs">
  <a href="#">...</a>
  <a href="#">...</a>
</div>
```

A little less verbose. No strong feeling if that breaks convention or there are objections.

docs/components/breadcrumbs.html
```diff
@@ -0,0 +1,61 @@
...
+        <h2 class="dialtone-title__section">Examples</h2>
+    </header>
+    <div class="dialtone-example" data-example="yes" data-code="yes">
+        <div class="dialtone-example__example dialtone-section16">
+            <div class="d-padding16">
+                <a href="#" class="d-breadcrumb">Root</a>
+                <a href="#" class="d-breadcrumb">Section</a>
+                <a href="#" class="d-breadcrumb">Section</a>
+                <a href="#" class="d-breadcrumb">Section</a>
+                <a href="#" class="d-breadcrumb--selected">Current Page</a>
```
tedgoas-dialpad, line 32: I don't think the current page should be linked.

lib/build/less/components/breadcrumbs.less
```diff
@@ -0,0 +1,90 @@
...
+//
+//  TABLE OF CONTENTS
+//  • BASE STYLE
+//  • INVERTED STYLE
+//
+//  ============================================================================
+//  $  BASE STYLE
+//  ----------------------------------------------------------------------------
+.d-breadcrumb {
+    box-sizing: border-box;
```
tedgoas-dialpad, line 16: We have to declare this each time? This isn't part of a CSS reset cascade or something?

lib/build/less/components/breadcrumbs.less
```diff
@@ -0,0 +1,90 @@
...
+
+    display: inline-flex;
+    position: relative;
+    .d-padding1();
```
tedgoas-dialpad, line 20: Should we be using variables here? Eg. `padding: @d-padding1;` or similar? Folks have advised against using mixins like this in the past, though I'm not sure how we feel about it here.

lib/build/less/components/breadcrumbs.less
```diff
@@ -0,0 +1,90 @@
...
+    .d-padding-right12();
```
tedgoas-dialpad, line 21: I dig the approach of adding the `/` this way, but adding padding to the link to make room for it makes the perceivable link longer than it should be. 

Screen Recording 2020-07-14 at 1 02 21 PM

lib/build/less/components/breadcrumbs.less
tedgoas-dialpad: If we're going the mixin route, do we need the parenthesis? Would `.d-padding1;` be acceptable?

tedgoas-dialpad (changes requested):
Two _actual_ changes suggested, rest are questions.

docs/components/breadcrumbs.html
```diff
@@ -0,0 +1,61 @@
...
+        <p class="dialtone-copy__normal">Breadcrumbs are always treated as secondary and should not entirely replace the primary navigation. They shouldn’t be used for products that have single level navigation because they create unnecessary clutter.</p>
+    </section>
+</section>
+<section class="dialtone-section16">
+    <header class="dialtone-section2">
+        <h2 class="dialtone-title__section">Examples</h2>
+    </header>
+    <div class="dialtone-example" data-example="yes" data-code="yes">
+        <div class="dialtone-example__example dialtone-section16">
+            <div class="d-padding16">
```
dbecherdialpad, line 27: We should use a `<nav>` and ensure the examples have the correct aria attributes https://www.aditus.io/patterns/breadcrumbs/

docs/assets/css/dialtone.css
```diff
@@ -8288,6 +8288,176 @@ legend.d-label--normal {
...
+  color: #A4A7AE;
+  color: var(--ash-darkest);
+}
+.d-breadcrumb--inverted--selected {
+  display: inline-flex;
+  position: relative;
+  box-sizing: border-box;
+  padding: 0.1rem;
+  padding-right: 1.2rem;
+  color: #6C6E75;
```
dbecherdialpad, line 8417: I thought we fixed the double-output problem? 🤔

lib/build/less/components/breadcrumbs.less
dbecherdialpad: We don't set `box-sizing` globally in Dialpad.

lib/build/less/components/breadcrumbs.less
dbecherdialpad: The `.d-padding1;` syntax is deprecated in less, it will need to be `.d-padding1();` in the future.

lib/build/less/components/breadcrumbs.less
dbecherdialpad: Will screen readers read the "/" separator? Looking at it briefly I think they will: https://thatdevgirl.com/blog/before-after-accessibility

docs/components/breadcrumbs.html
tedgoas-dialpad: +1 `<nav aria-label="breadcrumb">`

PR #95, opened 2020-07-27
Convert Dialtone from Jekyll >> 11ty

hynes-dialpad:
This converts the Dialtone website over to Eleventy from Jekyll.


tedgoas-dialpad (approved):
This looks good to me. I'm sure we can improve things here and there, but let's make a clean break from jekyll to 11ty and improve from there!

PR #102, opened 2020-08-20
Tabs

hynes-dialpad:
This add tab styles into Dialtone. I've provided HTML ARIA role and accessibility items to tabs as well.

image

## To-do:
- [x] Finish tab description
- [x] Add ARIA / Accessibility information
- [x] Simple JS to hook up examples 😎


docs/_includes/head.html
```diff
@@ -38,6 +38,9 @@
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
     <link rel="stylesheet" type="text/css" href="/{{ site.paths.css }}/dialtone-docs.css?{{ site.time | date: '%s%N' }}"/>
 
+    <!-- Webfonts -->
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,400;0,700;1,400;1,700&display=swap" rel="stylesheet">
```
tedgoas-dialpad: What is this for?

docs/assets/less/dialtone-docs.less
```diff
@@ -40,6 +40,10 @@
 @nav__background-color--l:      lightness(@nav__background-color);
 
 @define-font-path:              "../fonts/MarkPro/";
+@define-font-name:              "Roboto", @font-family--sans;
```
tedgoas-dialpad: Should we be doing this in _this_ PR?

lib/build/less/components/tabs.less
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    &:after {
+        content: "";
+        position: absolute;
+        right: 0;
+        bottom: 0;
+        left: 0;
+        height: @spacing4;
+        border-radius: @spacing4 @spacing4 0 0;
```
tedgoas-dialpad: It's a personal preference, but this part is throwing me off visually. I'm just like... "why?"

lib/build/less/components/tabs.less
```diff
@@ -0,0 +1,178 @@
...
+        margin-right: 0;
+    }
+
+    &:after {
+        content: "";
+        position: absolute;
+        right: 0;
+        bottom: 0;
+        left: 0;
+        height: @spacing4;
```
tedgoas-dialpad: Tried halving this to 2px and no border radius 🤷‍♀️ Again, personal preference but I like it. Feels more crisp.

Screen Shot 2020-08-20 at 4 25 24 PM

lib/build/less/components/tabs.less
```diff
@@ -0,0 +1,178 @@
...
+
+    border: 0;
+    border-radius: @spacing4 @spacing4 0 0;
+    background-color: fade(@white,0%);
+    font: inherit;
+    line-height: 1;
+    color: inherit;
+    fill: currentColor;
+    transition: all @transition--duration60 @transition--cubic-inOut;
+
```
tedgoas-dialpad, line 64: We may want a `cursor: pointer` here or the dialtone equivalent. I wasn't getting any visual indication these are clickable on `:hover`.

lib/build/less/components/tabs.less
```diff
@@ -0,0 +1,178 @@
...
+        z-index: 2;
+    }
+
+    &:hover {
+        color: var(--text-color--dark);
+        background-color: @ash;
+    }
+
+    //  Turn off animations if someone doesn't want them.
+    @media (prefers-reduced-motion) {
```
tedgoas-dialpad, line 93: I wonder if we could put this in a more global location so that we don't have to redefine it for each component.

docs/components/tabs.html
```diff
@@ -0,0 +1,120 @@
...
+                {% assign desc = i.desc %}
+                <tr>
+                    <th scope="row">.{{ class }}</th>
+                    <td>{{ desc }}</td>
+                </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+    </div>
+</section>
```
tedgoas-dialpad, line 210: Do we want to add a section for responsive? Eg. What the "More" menu is invoked?

docs/components/tabs.html
```diff
@@ -0,0 +1,120 @@
+---
+layout: page
+title: Tabs
+description: TBD
+---
+<section class="dialtone-section8">
```
tedgoas-dialpad, line 6: Do we want to include a Do's and Don'ts section here to advise folks when to use tabs and when "just add another tab" isn't appropriate?

tedgoas-dialpad (changes requested):

lib/build/less/components/tabs.less
hynes-dialpad: Possibly. I'd rather look at doing a universal selector like that in a separate PR. For now, we can at least do it here.

docs/components/tabs.html
hynes-dialpad: I'm going to push this to a future iteration as it will require some JS work.

docs/components/tabs.html
hynes-dialpad: Probably, but we need to do this for all components. I'd rather tackle it then.

dbecherdialpad (approved):

tedgoas-dialpad:
@hynes-dialpad We spoke about tab width. (I think) we agreed to make the tabs less wide by default and adding modifiers to make them wider and include an underline. Something like:

`d-tablist`(Narrow width, no underline)
`d-tablist d-tablist__wide`
`d-tablist d-tablist__underline`

Issue #101, opened 2020-08-20
Utility Classes


tedgoas-dialpad:
- [x] **🔥 Display**: We have `display: grid` and `display: flex` covered, but what about `inline`, `inline-block`, `block`, `none` and the `table` properties?
- [x] **🔥 Flexbox**: Add all of flexbox's additional properties for alignment and spacing. Honestly I think [Stacks](https://stackoverflow.design/product/base/grid/#flex-classes) does a great job here.
- [ ] **Floats**
- [ ] **Cursors**: [Cursors](https://tailwindcss.com/docs/cursor) and [pointer events](https://stackoverflow.design/product/base/interactivity/).
- [x] **🍀 Lists**
- [ ] **Object Fit & Position**
- [x] **🍀 Overflow**
- [x] **🍀 🔥 Positioning**: Including [top, right, bottom, left](https://tailwindcss.com/docs/top-right-bottom-left)
- [x] **🍀 🔥 Text Align**
- [x] **🍀 Text Decoration**
- [x] **🍀 Text Transform**
- [ ] **Transitions**
- [ ] **Visibility**
- [x] **🍀 Whitespace**
- [x] **🍀 Word Break**

🔥  - high impact
🍀 - in code review

---

- [ ] **Responsive Design**: We should make everything responsive and [do it with utility classes](https://tailwindcss.com/docs/responsive-design#app). Do we have breakpoint variables defined?
- [ ] **Screen-Readers**: [Very helpful](https://tailwindcss.com/docs/screen-readers)!

PR #107, opened 2020-08-27
Add flexbox utility classes

tedgoas-dialpad:
This adds more utility classes for common flexbox properties:

- [x] Justify Content
- [x] Align Items
- [x] Align Content
- [x] Align Self
- [x] Flex Flow (docs only)
- [x] Flex Direction
- [x] Flex Wrap
- [x] Update example blocks for `flex`, `flex-grow`, and `flex-shrink`

Part of #101


hynes-dialpad (changes requested):
A few more items we should add:

- Flex Direction
- Flex Wrap
- `unset` options for each utility class.

Here's a Less Flex utility file that used to be in Dialtone that we could re-add: https://github.com/dialpad/dialtone/blob/v2.0.0/lib/less/utilities/flex.less

lib/build/less/utilities/flex.less
```diff
@@ -70,8 +75,72 @@
...
 //      Designate a child container's position.
 //  ----------------------------------------------------------------------------
 .d-order--first                         { order: -1; }
 .d-order--last                          { order: 1; }
+
+//  --  FLEX FLOW
+//      This is another shorthand property for flex-direction and flex-wrap.
+//      Default value is "row nowrap".
+//  ----------------------------------------------------------------------------
+.d-row-wrap              { flex-flow: row wrap !important; }
```
hynes-dialpad: These utilities don't follow the same pattern as all the other utilities. These should probably be `d-ff-row-nowrap`, etc.

tedgoas-dialpad: Good call, updated.

lib/build/less/utilities/flex.less
```diff
@@ -22,7 +23,12 @@
 //  ============================================================================
 //  $   FLEXBOX UTILITIES
 //  ============================================================================
-//  --  FIXED CELL WIDTHS
+//  --  FLEXBOX
+//  ----------------------------------------------------------------------------
+.d-flex             { display: flex; }
```
tedgoas-dialpad: @hynes-dialpad Adding utility classes here, but leaving `display: flex;` inside `.d-flex-columns` for now since I don't know what will break if we remove it.

tedgoas-dialpad:
@hynes-dialpad 

> A few more items we should add:
> * Flex Direction
> * Flex Wrap

We have classes for `flex-flow`, which is shorthand for `flex-direction` and `flex-wrap`. Seems to cover all the cases. Good enough?

dbecherdialpad (approved):

hynes-dialpad (approved):

PR #109, opened 2020-08-27
Add utility classes for Display

tedgoas-dialpad:
Adds variables and utility classes for the `display` property.

- [x] block
- [x] inline-block
- [x] inline
- [x] table
- [x] none
- [x] unset

How do we want to handle `flex` and `grid`, as they're already defined elsewhere?

---

Do we want to include `inline-flex`, `inline-grid`, `contents`, `table-caption`, `table-cell`, `table-column`, `table-column-group`, `table-footer-group`, `table-header-group`, `table-row-group`, `table-row`, or `flow-root`?


docs/_data/display.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  {
+    "name": ".d-none",
+    "variable": "@d-none",
+    "output": "display: none;"
+  },
+  {
+    "name": ".d-unset",
+    "variable": "@d-unset",
+    "output": "display: unset;"
```
dbecherdialpad: imo these names are a bit too terse, `.d-none` sounds like it could refer to a lot of things

dbecherdialpad (commented):
We probably want to include inline-flex and inline-grid, maybe as modifiers for the existing flex and grid utilities. We would need all the table ones if we include `display: table` but I'm not convinced we need to include that one.

lib/build/less/utilities/display.less
```diff
@@ -0,0 +1,21 @@
...
+//  visit https://dialpad.design/
+//
+//  TABLE OF CONTENTS
+//  • Generic Loader
+//
+//  ============================================================================
+//  @   DISPLAY PROPERTIES
+//  ----------------------------------------------------------------------------
+
+.d-block          { display: @d-block !important; }
```
hynes-dialpad: These utility classes should include the CSS abbreviation even though it would be a little repetitious. Instead of `d-block`, let's do `d-d-block`.

These classes used to be in Dialtone. You can view them here: https://github.com/dialpad/dialtone/blob/v2.0.0/lib/less/utilities/layout.less

lib/build/less/variables/display.less
```diff
@@ -0,0 +1,20 @@
...
+//  For further documentation of these and other classes,
+//  visit https://dialpad.design/constants/display
+//
+//  TABLE OF CONTENTS
+//  • Display variables
+//
+//  ============================================================================
+//  $   DISPLAY
+//  ----------------------------------------------------------------------------
+@d-block:         block;
```
hynes-dialpad: Why do we need variables for display?

hynes-dialpad (changes requested):

docs/_data/display.json
tedgoas-dialpad: Yea, Hynes made a similar comment. I'll update.

tedgoas-dialpad:
> We would need all the table ones if we include display: table but I'm not convinced we need to include that one.

[I removed all `table` classes](https://github.com/dialpad/dialtone/pull/109/commits/d244e5d0ed4bf39bffaee2e3559efaad1d6ac50f) because any problems they'd solved would likely be solved by using Flexbox, Grid, or semantic table markup.

cc @dbecherdialpad

lib/build/less/variables/display.less
tedgoas-dialpad: No, we really don't. I put them in so folks wouldn't have to type `block` or `inline` but it doesn't save any time or fix any issues. I'll remove.

dbecherdialpad (approved):

hynes-dialpad (approved):

PR #111, opened 2020-09-03
Increase color contrast of disabled radios and checkboxes

tedgoas-dialpad:
Natayi pointed out how our `disabled` radios and checkboxes have poor color contrast.

Apparently certain disabled UI elements don't need to pass AA or AAA contrast to be considered accessible. When a something is hard to read, users don’t bother with it, which is the intent of a disabled button or label.

> Text or images of text that are part of an inactive user interface component, that are pure decoration, that are not visible to anyone, or that are part of a picture that contains significant other visual content, have no contrast requirement.

However I'd argue you still need to see _what's_ disabled.

For this reason, I've gone with `ash-darkest` as a base color for disabled states. It doesn't quite pass AA, but it doesn't need to. `slate-lightest` is the next step darker and _does_ pass contrast, but it's much closer to the color of an enabled input + label.


tedgoas-dialpad:
Preview [checkboxes](https://deploy-preview-111--dialtone.netlify.app/components/checkboxes/) and [radios](https://deploy-preview-111--dialtone.netlify.app/components/radios/).

hynes-dialpad (approved):

PR #113, opened 2020-09-10
Typography utility classes

tedgoas-dialpad:
This PR adds immutable classes from our existing typography mixins and creates immutable utility for missing typography CSS properties (listed below). Dialtone's `/utilities/typography/` docs are also updated to match the style and format of other doc pages.

**Update:**
- [x] Make immutable classes from existing mixins.
- [x] Update docs and add examples to existing 4 pages.

**Add:**
- [x] Styles / Font style
- [x] Styles / Text decoration
- [x] Styles / Text-transform
- [x] Styles / Lists
- [x] Layout / Text align
- [x] Layout / Truncate
- [x] Layout / Whitespace
- [x] Layout / Word break


lib/build/less/utilities/typography.less
```diff
@@ -41,6 +41,13 @@ if((@define-font-face = true), each(@define-font-family, {
 .d-font-family--inherit         { font-family: inherit; }
 .d-font-family--unset           { font-family: unset; }
 
+// Utility classes
+.d-font-family-custom          { .d-font-family--custom() !important; }
+.d-font-family-sans            { .d-font-family--sans() !important; }
```
dbecherdialpad: is `d-font-family-sans` needed?

dbecherdialpad (approved):

lib/build/less/utilities/typography.less
tedgoas-dialpad: I think so. Why wouldn't we want it?

dbecherdialpad: I mean, I can't think of anywhere in the app where we want to use the system sans-serif font specifically instead of mark/roboto but if that's desirable then it's fine with me.

tedgoas-dialpad: That's a fair point. I didn't immediately realize our default is `font-family--custom`. I wonder if it would make sense to kill that and add mark/roboto `d-font-family-sans` and use that as our default. Thoughts @hynes-dialpad?

hynes-dialpad: There are times right now where for legibility reasons, we do use `d-ff-sans`. It's rare, but still used. We shouldn't be restating `d-ff-custom` as much as we go though.

hynes-dialpad (approved):

PR #117, opened 2020-09-21
(Documentation): fix example for stacks

vipul-b:


hynes-dialpad (approved):
Thanks for starting this PR @vipuldialpad

PR #119, opened 2020-09-24
Add utility classes for the CSS overflow property

tedgoas-dialpad:
This works towards #101


vipul-b:
I think we can add utility classes that build on top of this one.

A common text-related utility class would, for example, provide support for truncating long lines of text and show ellipses at the end and there are similar use-cases we might want to capture in CSS classes or mixins.

tedgoas-dialpad:
@vipuldialpad Yup! There are a few classes for text truncation in https://github.com/dialpad/dialtone/pull/113

hynes-dialpad (approved):

PR #134, opened 2020-10-24
Add background patterns

hynes-dialpad:
## Description
This adds background patterns into Dialtone as background images. This achieved by converting SVGs into base64 images.

image

Because the `background-image` property for the pattern might conflict with a linear-gradient if paired together, I'm proposing to include background patterns in an absolutely positioned `div`. This approach avoids potential `background-image` conflicts as well allows the empty pattern `div` to fill 100% of its parent container without needing a height attached.

This will mean that any element utilizing a background pattern will need to have `d-p-relative` (`position: relatiave;`) applied to it so the pattern sits properly in there. Also it's recommended to apply `d-pl24` to the parent wrapper to push content away from the pattern.

## Pull Request Checklist


MikaDialpad (approved):
I don't normally review dialtone PRs, but was that class change on all those hundreds of files intentional? :D

I think this approach will work for us. It adds a bit more responsibility to the engineer to make it work, but like you said, using multiple background wouldn't really be possible othwerise. I was also thinking if we could use a pseudo element within the mixin to create the pattern div, but that could also conflict if the container had such defined. Anyways, these patterns are pretty rarely used, so it shouldn't be too much trouble to implement them as instructed. Thanks for getting this done Josh!

hynes-dialpad:
A pseudo element is an interesting idea @MikaDialpad. 🤔  It wouldn't be hard to use that approach, and probably would be easier to maintain. Let me test it out quickly today.

docs/assets/less/dialtone-docs.less
```diff
@@ -756,14 +756,6 @@ body {
     .d-margin-bottom0();
     font-size: @fs16;
     line-height: @lh-loose;
-
```
tedgoas-dialpad, line 760: Will this break other lists on the docs site?

tedgoas-dialpad (changes requested):
I think we need to document when these are useful and how they should be used. This current page should tell us:

* When they should be used. What components or product areas are they required. Where are they optional? Where should they be avoided?
* How they should be used. Does each pattern mean something specifically or are the patterns random? How should we assign patterns.

The code looks Ok, I’m just not sure when we should be reaching for these.

docs/assets/less/dialtone-docs.less
hynes-dialpad: It shouldn't. It should default back to Dialtone.

hynes-dialpad:
@TedGoas 

> - When they should be used. What components or product areas are they required. Where are they optional? Where should they be avoided?
> - How they should be used. Does each pattern mean something specifically or are the patterns random? How should we assign patterns.

They should be used **only** for Department and Call Center labels as stated in the first sentence of ["Background Patterns" page](https://deploy-preview-134--dialtone.netlify.app/utilities/visual/background-patterns/): "Dialtone provides various patterns to be used for Department and Call Centers." They are randomly assigned with a color for a department of call center. Patterns do not contain meanings. These are being added for accessibility reasons to help with with better differentiate colors. This is also documented on the [Resources > Patterns page](https://deploy-preview-134--dialtone.netlify.app/icons/patterns/):

> The following patterns used for Department and Call Centers, paired with various colors to help with low-color or color-blindness issues.

tedgoas-dialpad:
I completely missed those parts on both pages. IMO this part of the page is doing a little too much:
Screen Shot 2020-10-28 at 11 09 43 AM

Other pages don't include usage guidelines in the page title area. I think we should move some of this paragraph into its own section called "Guidelines", "Usage", or something like that.

hynes-dialpad:
@tedgoas-dialpad  I've tried to clarify the documentation page a bit more. It probably needs another pass, but I'm inclined to push this out and polish the documentation page in future updates.

tedgoas-dialpad:
Agree, I think this is much better and we can improve further down the road.

tedgoas-dialpad (approved):

PR #147, opened 2020-12-01
Added my name to contributors list

nabasolo-dialpad:
Trying out setting up Dialtone on my machine, so I added my name to the contributors list to test out my understanding of the process.

@TedGoas , I'm not sure why it's saying I edited some of the css files, I didn't touch them at all. Is there a way to only approve that one change I did and not the rest?


hynes-dialpad:
@abasolo-dialpad It's fine. The CSS files are compiled with every build. These CSS files are only used in the Documentation site.

tedgoas-dialpad (approved):
💪 Nice work!

PR #174, opened 2021-02-02
Issue170 page toc overflow

### LINKED ISSUE #170: Bug: Page content overflows TOC
Body content overflows into table of tables.

On Chrome:
Screen Shot 2021-02-01 at 12 28 23 PM

dialpad-drew:
## Description
Introduces a new grid layout class called `.d-gl-aside` to accommodate "[sidebar] [main] [sidebar]" layouts. It also removes dozens of the explicit character unit width declarations (`d-w75ch`), which caused an issue in which main content would overflow beyond its parent container. This should fix #170 

Because of how CSS grid targets, I also needed to move the TOC `aside` to be a sibling of `main` and `nav`.

In an effort to improve responsiveness, this PR also adds `md:d-d-block` to the element with the CSS Grid class. This makes the `main` content area a full-width block element at 640px and below.

In truth, I would prefer _not_ to stop using CSS Grid at certain breakpoints and instead just change the `grid-template-areas`, but I can't figure out how to do so with this immutable utility class approach.  If there's a better way to do this, I'm all-ears.

### Before
Screen Shot 2021-02-01 at 3 34 32 PM

### After
Screen Shot 2021-02-02 at 11 30 39 AM

## Pull Request Checklist

 - [X] Review the [contribution guidelines](https://github.com/dialpad/dialtone/blob/staging/.github/CONTRIBUTING.md).
 - [X] Ensure all `gulp` scripts successfully compile.
 - [X] Update, remove, or extend all affected documentation.
 - [X] Provide a description what is being changed, extended, or removed.
 - [X] Link to any issue(s) this request is resolving.
 - [X] Request a review from Drew Chandler, Joshua Hynes, or Ted Goas.


hynes-dialpad:
@dialpad-drew I'm not sure we would use this class beyond this documentation site. Anywhere else in the product that we have a similar three column layout, the right sidebar acts more as a drawer that opens, closes, or expands over content. I wonder if the responsive `display` classes would solve this problem better. 🤔

dialpad-drew:
> @dialpad-drew I'm not sure we would use this class beyond this documentation site. Anywhere else in the product that we have a similar three column layout, the right sidebar acts more as a drawer that opens, closes, or expands over content. I wonder if the responsive `display` classes would solve this problem better. 🤔

@hynes-dialpad  Gotcha and fair point. I think the commit I just pushed accomplishes essentially the same thing I was originally going for without adding any additional classes. I'm not nearly as familiar with these utility classes as you are, but I'm not sure I see a way to do this using `grid-area` and `grid-template-areas` without adding more classes or nesting the TOC into the main content, so I pulled out those classes and went with a more generic 12-col approach. 🤷‍♂️ 

I started with a [2-col][8-col][2-col] approach, but the main content well was slightly too large and the left navigation was pretty tight. Because `position: fixed` takes that left nav out of the CSS flow, it doesn't wrap and the layout just wasn't gracefully fluid.

I also tried [3-col][6-col][3-col] and thought it provided too much space for the TOC, especially on larger viewports. So, I settled on an asymmetrical [3-col][7-col][2-col]. It gives you some of that left nav breathing room without the explicit width.
Screen Shot 2021-02-03 at 11 06 07 AM

Also, speaking of explicit width: was there a reason the left nav had `width: 28rem` (see `nav.html:2`)? I had to remove that because it was cramping the layout's fluidity by overflowing into the main content.

hynes-dialpad:
@dialpad-drew My only comment on this is that I feel the left sidebar should have a max width. Is there a layout where we could something like:

```
[ Sidebar minmax(min-content, 30rem(??))] [ Main 1fr ] [TOC minmax(min-content, 21.6rem??)]
```

This is also something we can clean up later as we deal with the making the header responsive.

dialpad-drew:
> @dialpad-drew My only comment on this is that I feel the left sidebar should have a max width. Is there a layout where we could something like:
> 
> ```
> [ Sidebar minmax(min-content, 30rem(??))] [ Main 1fr ] [TOC minmax(min-content, 21.6rem??)]
> ```
> 
> This is also something we can clean up later as we deal with the making the header responsive.

I'm sure there could be. I'm confused, though, because the instructions were "no more CSS."

dialpad-drew:
> @dialpad-drew My only comment on this is that I feel the left sidebar should have a max width. Is there a layout where we could something like:
> 
> ```
> [ Sidebar minmax(min-content, 30rem(??))] [ Main 1fr ] [TOC minmax(min-content, 21.6rem??)]
> ```
> 
> This is also something we can clean up later as we deal with the making the header responsive.

~I'm sorry, man, but I'm not sure what you want. It kinda feels like we're going back and forth here. Since it's going to take you 90 seconds versus me a half-hour, do you mind just committing and pushing the change you hope to see?~ What I pushed isn't responsive whatsoever, but it fixes the sidebar width.

PR #201, opened 2021-03-10
Dialtone V6 Component Tooltip Implementation

### LINKED ISSUE #199: [CSS] Component: Tooltip
**Acceptance Criteria**

- Update the tooltip component Dialtone classes

**Resources**

Design: https://www.figma.com/file/zP2uyVxAg4notGNZwW64mW/Components-(Dialtone-6.0)?node-id=4518%3A13082

ryandialpad:
# Dialtone V6 Component Tooltip Implementation

## :hammer_and_wrench: Type Of Change

- [ ] Fix
- [x] Feature
- [x] Refactoring
- [ ] Documentation

## :book: Description

Updated tooltips component classes
Minor updates to documentation

## :bulb: Context

To implement V6 Tooltips components
Resolves #199 

## ✏️  Pull Request Checklist

## :crystal_ball: Next Steps

Update tooltips documentation
Scope work for tooltips Vue implementation

## ❌  Breaking Changes

* `d-tooltip__{DIRECTION_1}-{DIRECTION_2}` is now `d-tooltip__arrow--{DIRECTION_1}-{DIRECTION_2}`
* must use `d-tooltip--hover`, `d-tooltip--show` and `d-tooltip--hide` to control visibility
* hover mixin `.d-tooltip-hover()` removed

### Using Hide / Show

```vue
<template>
  <div class="`d-tooltip d-tooltip__arrow--{dir1}-{dir2} d-tooltip--{visibility}`">
    <slot>
      <span>{{ message }}</span>
    </slot>
  </div>
</template>
```

### Using Hover

```vue
<template>
  <div class="d-tooltip--hover">
    <slot />
    <div class="`d-tooltip d-tooltip__arrow--{dir1}-{dir2}`">
      <slot #message>
        <span>{{ message }}</span>
      </slot>
    </div>
  </div>
</template>
```


lib/less/components/tooltips.less
```diff
@@ -1,28 +1,84 @@
...
+    .d-tooltip__content {
+        transform: translate(0,0);
+        opacity: 1;
+        visibility: visible;
+        transition:
+            visibility 0s 0s,
+            transform 200ms var(--ttf-in-out) 10ms,
+            opacity 200ms var(--ttf-in-out) 10ms;
+    }
+}
```
ryandialpad: @dbecherdialpad @braddialpad @dialpad-drew My CSS / LESS skills are a little rusty, can anyone of you think of a better way to write these selectors?

ryandialpad:
@dialpad-drew @hynes-dialpad I was wondering if I could get a quick once over on the tooltips component class changes?

lib/less/components/tooltips.less
```diff
@@ -1,28 +1,84 @@
...
 
+//  ============================================================================
+//  $   CONTENT STYLE
+//  ----------------------------------------------------------------------------
+.d-tooltip__content {
```
dialpad-drew: Can we make this selector's padding 8px on all sides instead of 6px on the bottom? That'll match the Figma UI.

dialpad-drew (commented):
Any idea why some of these examples' widths are collapsing?

Screen Shot 2021-03-10 at 3 43 20 PM
Screen Shot 2021-03-10 at 3 43 25 PM

dbecherdialpad:
So I have a little problem with the pure CSS tooltip implementation which prevents us from using these styles in the actual product:

1. Showing the tooltip on `:hover` is not accessible.
2. Tooltips should be rendered at the root of the DOM and positioned to the target element (aka "teleport"). This is because the tooltip will be clipped to the bounding box if it's in a scroll container, and sometimes get resized to the parent element as noted here.
3. If there are many tooltips on the page they should lazy render only when needed.

IMO we should change the dialtone tooltip implementation to be visual styles only and implement the rest of the desired functionality in the Vue component. Currently we use the `v-tooltip` directive for this which is not very good.

dialpad-drew:
> 1. Showing the tooltip on `:hover` is not accessible.

This is a good point. However, why wouldn't we just expose the Tooltip on `:focus` as well? This seems inline with how [Heydon Pickering](https://inclusive-components.design/tooltips-toggletips/) talks about it, and how [Carbon](https://www.carbondesignsystem.com/components/tooltip/code) does their **Icon tooltip** and **Definition tooltip**.

dbecherdialpad:
>. However, why wouldn't we just expose the Tooltip on :focus as well?

Imagine the following scenario:

1. A button has a tooltip that shows on hover or focus.
2. You hover over the button and click on it.
3. You move the mouse away from the button.
4. The tooltip is stuck open until you click off the button because it's still focused.

I'm not aware of a pure CSS solution to this unless `:focus-visible` solves it.

Been a while since I looked into it, but I also think you'll need to update `aria-hidden` and `aria-describedby` on the tooltip and anchor elements respectively which requires JS anyway. Also in that Carbon example, the focus moves into the tooltip when it opens which would then close the tooltip with a pure CSS solution (that's really what we call a popover though).

lib/less/components/tooltips.less
ryandialpad: Great catch!

🤦  My apologies, I wanted to get an initial review as quick as possible to discuss the overall implementation.

ryandialpad:
> So I have a little problem with the pure CSS tooltip implementation which prevents us from using these styles in the actual product:
> 
> 1. Showing the tooltip on `:hover` is not accessible.
> 2. Tooltips should be rendered at the root of the DOM and positioned to the target element (aka "teleport"). This is because the tooltip will be clipped to the bounding box if it's in a scroll container, and sometimes get resized to the parent element as noted here.
> 3. If there are many tooltips on the page they should lazy render only when needed.
> 
> IMO we should change the dialtone tooltip implementation to be visual styles only and implement the rest of the desired functionality in the Vue component. Currently we use the `v-tooltip` directive for this which is not very good.

Good call, I have since removed the `hover` class just leaving the `show` and `hide` classes which can be used to toggle the tooltips visibility.

ryandialpad:
> Any idea why some of these examples' widths are collapsing?
> 
> Screen Shot 2021-03-10 at 3 43 20 PM
> 
> Screen Shot 2021-03-10 at 3 43 25 PM

No idea currently, will investigate though

lib/less/components/tooltips.less
ryandialpad: @dialpad-drew The current styles use rem, did we want to continue to use rems? or follow the design and go with pixels?

`padding: var(--su8) var(--su8) var(--su6);`

ryandialpad: Same goes with `border-radius` which is `var(--su4)`

ryandialpad:
```less
  --su4: 0.4rem;
  --su6: 0.6rem;
  --su8: 0.8rem;
```

ryandialpad: Finally for line height we have the following builtin options

```less
--lh-tight: 1.25; // 15px @ 12px font
--lh-normal: 1.5; // 18px @ 12px font
```

ryandialpad:
@dialpad-drew Is there a max width that you have in mind for the tooltips? or shall we remove the existing max width?

I would personally opt against defining a max width in the base component

lib/less/components/tooltips.less
ryandialpad: Found another `border: var(--su6) solid transparent;`

ryandialpad: Basically everything is using variables which map to rems

dialpad-drew:
> > . However, why wouldn't we just expose the Tooltip on :focus as well?
> 
> Imagine the following scenario:
> 
> 1. A button has a tooltip that shows on hover or focus.
> 2. You hover over the button and click on it.
> 3. You move the mouse away from the button.
> 4. The tooltip is stuck open until you click off the button because it's still focused.
> 
> I'm not aware of a pure CSS solution to this unless `:focus-visible` solves it.
> 
> Been a while since I looked into it, but I also think you'll need to update `aria-hidden` and `aria-describedby` on the tooltip and anchor elements respectively which requires JS anyway. Also in that Carbon example, the focus moves into the tooltip when it opens which would then close the tooltip with a pure CSS solution (that's really what we call a popover though).

Ultimately, here are my concerns about pulling the functionality from the CSS component:

1. It doesn't work out of the box and we're documenting is essentially a static visual styles, does Tooltip still meet the criteria for being considered a true, first-class CSS component?
2. Who's responsible for updating all the Tooltip CSS implementations to Vue when we break them when updating projects to DT6? If it's us, then I wanna make sure we trap that as part of the implementation feature.

dialpad-drew:
> @dialpad-drew Is there a max width that you have in mind for the tooltips? or shall we remove the existing max width?
> 
> I would personally opt against defining a max width in the base component

Honestly, I have no opinion. I haven't really decomposed this component and, unless it's creating problems, would prefer the dimensions to mirror what's live today: https://dialpad.design/components/tooltips/

lib/less/components/tooltips.less
dialpad-drew: Static pixels = 👎 
REM = 👍 

I didn't touch the border-radius or line-height in Figma. Just the font, font-color, and tooltip background color.

ryandialpad: @dialpad-drew For the padding, should we keep it the same on all sides and just go with `--su8`?

ryandialpad:
> > @dialpad-drew Is there a max width that you have in mind for the tooltips? or shall we remove the existing max width?
> > I would personally opt against defining a max width in the base component
> 
> Honestly, I have no opinion. I haven't really decomposed this component and, unless it's creating problems, would prefer the dimensions to mirror what's live today: https://dialpad.design/components/tooltips/

They weren't required and as such were removed

ryandialpad:
One issue that I had noticed, it seems that the tooltip is pushed outside of the div when it is not attached to an element like the previous implementation.

Screen Shot 2021-03-11 at 2 43 22 PM

dialpad-drew:
@ryandialpad luck of the draw with tooltips being the first component, but these updates broke the tooltips in the top nav. 

Expected UI:
Screen Shot 2021-03-11 at 2 46 26 PM
Screen Shot 2021-03-11 at 2 46 25 PM
Screen Shot 2021-03-11 at 2 46 23 PM

vs. 

What's in this branch:
Screen Shot 2021-03-11 at 2 50 27 PM

ryandialpad:
Tooltip positioning is now fixed

Screen Shot 2021-03-11 at 4 54 14 PM

ryandialpad:
> @ryandialpad luck of the draw with tooltips being the first component, but these updates broke the tooltips in the top nav.
> 
> Expected UI:
> Screen Shot 2021-03-11 at 2 46 26 PM
> Screen Shot 2021-03-11 at 2 46 25 PM
> Screen Shot 2021-03-11 at 2 46 23 PM
> 
> vs.
> 
> What's in this branch:
> Screen Shot 2021-03-11 at 2 50 27 PM

@dialpad-drew Since we changed the css implementation, whatever uses the tooltip in our docs site is now broken. Will address the issue tomorrow. We will need JS to control the visibility of the tooltip.

ryandialpad:
@dialpad-drew Completed updates to the header, let me know what you think

docs/assets/js/header.js
```diff
@@ -0,0 +1,34 @@
+$(document).ready(function() {
```
ryandialpad: Since the tooltip component does not include a hover style, we need JS to toggle the style. The script below will handle swapping the visibility classes for each of header tooltips.

docs/assets/js/header.js
```diff
@@ -0,0 +1,34 @@
...
+    .mouseenter(() => showTooltip(headerGitRepoTooltip))
+    .mouseleave(() => hideTooltip(headerGitRepoTooltip));
+  
+  headerCodepenTemplateIcon
+    .mouseenter(() => showTooltip(headerCodepenTemplateTooltip))
+    .mouseleave(() => hideTooltip(headerCodepenTemplateTooltip));
+  
+  headerInvertColorsIcon
+    .mouseenter(() => showTooltip(headerInvertColorsTooltip))
+    .mouseleave(() => hideTooltip(headerInvertColorsTooltip));
```
ryandialpad: Triggering the visibility class swaps on the Icon rather than the container so that hovering over the tooltip does not keep it open.

dbecherdialpad:
@dialpad-drew 

> It doesn't work out of the box and we're documenting is essentially a static visual styles, does Tooltip still meet the criteria for being considered a true, first-class CSS component?

The same could be said for modals, tabs, and probably lots of other components. The "true" implementation is really the vue component which implements these styles.

> Who's responsible for updating all the Tooltip CSS implementations to Vue when we break them when updating projects to DT6? If it's us, then I wanna make sure we trap that as part of the implementation feature.

Ugh, yeah...there's plenty of places in the app that use the tooltip styles with the hover state. Is there any way to make those backwards compatible? Maybe an additional class to enable the old show-on-hover functionality?

> Is there a max width that you have in mind for the tooltips? or shall we remove the existing max width?

Tooltips should have a max width as long as we could override it, which shouldn't ever really be necessary. Imagine all this text on one line:

Screen Shot 2021-03-12 at 10 33 41 AM

ryandialpad:
> @dialpad-drew
> 
> > It doesn't work out of the box and we're documenting is essentially a static visual styles, does Tooltip still meet the criteria for being considered a true, first-class CSS component?
> 
> The same could be said for modals, tabs, and probably lots of other components. The "true" implementation is really the vue component which implements these styles.
> 
> > Who's responsible for updating all the Tooltip CSS implementations to Vue when we break them when updating projects to DT6? If it's us, then I wanna make sure we trap that as part of the implementation feature.
> 
> Ugh, yeah...there's plenty of places in the app that use the tooltip styles with the hover state. Is there any way to make those backwards compatible? Maybe an additional class to enable the old show-on-hover functionality?
> 
> > Is there a max width that you have in mind for the tooltips? or shall we remove the existing max width?
> 
> Tooltips should have a max width as long as we could override it, which shouldn't ever really be necessary. Imagine all this text on one line:
> 
> Screen Shot 2021-03-12 at 10 33 41 AM

@dialpad-drew Thoughts on a max-width to initially set?

@dbecherdialpad I am good to add back the third visibility style that I had created `d-tooltip--hover` to make the migration a little easier for some teams

dialpad-drew:
> @dialpad-drew Thoughts on a max-width to initially set?

@ryandialpad  I say we keep it the same that it is now. Looking at `5.x`, I believe it's `max-width: 21.6rem`

ryandialpad:
> > @dialpad-drew Thoughts on a max-width to initially set?
> 
> @ryandialpad I say we keep it the same that it is now. Looking at `5.x`, I believe it's `max-width: 21.6rem`

@dialpad-drew Thoughts on having the content center aligned by default or just leave it up to the user?

Drew: I don't think we should touch any of that right now and leave it as-is (centered). It's thorough thinking, but creeping our own scope.

> Also did you have a decision on the padding question above? Should we just use `su8` for each side instead of a mix of `su8` and a single `su6`?

Ryan: Tested locally, `su6` did not look good, went with `su8`
Drew: Sorry if I left you hanging -- thought we already covered this. `su8` on all sides please. 👍

dialpad-drew (approved):

ryandialpad:
@dialpad-drew @dbecherdialpad Added hover style back for now, component is now ready for review

dialpad-drew:
@ryandialpad this looks great! Thanks for being so thorough. 🙌 

I know I already approved this, but I'd like your opinion on one last thing: I know this component's CSS form isn't fully accessible (i.e. not visible on focus), but do you or @dbecherdialpad think it'd be worthwhile to at least include a voiceover readout on `:focus`? Maybe `aria-describedby` or something?

ryandialpad:
> @ryandialpad this looks great! Thanks for being so thorough. 🙌
> 
> I know I already approved this, but I'd like your opinion on one last thing: I know this component's CSS form isn't fully accessible (i.e. not visible on focus), but do you or @dbecherdialpad think it'd be worthwhile to at least include a voiceover readout on `:focus`? Maybe `aria-describedby` or something?

@dialpad-drew Thanks! My intention is to make sure that we ideally don't miss anything when transitioning to V6 and to think about the components holistically (css implementation + vue implementation) :)

In terms of the Voiceover readout, 100% we will ensure that the Vue implementation is as accessibly as possible, making sure that there is a readout when the tooltip is visible / focused (I need to read up further on the guidelines for accessible tooltip components).

dialpad-drew:
> In terms of the Voiceover readout, 100% we will ensure that the Vue implementation is as accessibly as possible, making sure that there is a readout when the tooltip is visible / focused (I need to read up further on the guidelines for accessible tooltip components).

Makes sense for the Vue component, but what about the HTML component we're working on here? If we think it's going to be used in the product, shouldn't we also provide a reasonably accessible experience as well?

ryandialpad:
> > In terms of the Voiceover readout, 100% we will ensure that the Vue implementation is as accessibly as possible, making sure that there is a readout when the tooltip is visible / focused (I need to read up further on the guidelines for accessible tooltip components).
> 
> Makes sense for the Vue component, but what about the HTML component we're working on here? If we think it's going to be used in the product, shouldn't we also provide a reasonably accessible experience as well?

@dialpad-drew When we write the docs for this css component we can include accessibility considerations :)

For this PR we are focused on the CSS implementation. Dialtone doesn't include any html components currently but it does include usage examples and in the future, links to the Vue implementations.

dialpad-drew:
> > > In terms of the Voiceover readout, 100% we will ensure that the Vue implementation is as accessibly as possible, making sure that there is a readout when the tooltip is visible / focused (I need to read up further on the guidelines for accessible tooltip components).
> > 
> > 
> > Makes sense for the Vue component, but what about the HTML component we're working on here? If we think it's going to be used in the product, shouldn't we also provide a reasonably accessible experience as well?
> 
> @dialpad-drew When we write the docs for this css component we can include accessibility considerations :)
> 
> For this PR we are focused on the CSS implementation. Dialtone doesn't include any html components currently but it does include usage examples and in the future, links to the Vue implementations.

Gotcha. I created issue #203 for this component to remind us to revisit this (maybe after we do the Vue counterpart?). You're right in that we aren't offering "HTML components," but our examples should probably still show an accessible implementation. 

Let's keep track of these as we go and revisit. :)

PR #222, opened 2021-03-17
V6 component link docs

dialpad-drew:
## Description
- Re-styles Link documentation
- Makes a couple style tweaks to the `codeWellHeader` and `codeWellFooter` shortcodes to smooth out some of the dark-background example styles
- Changes `--warning-color-focus` from `--yellow-200` to `yellow-400`. I think `--yellow-200` will be inaccessible and I didn't notice it until re-styling the page.
- Solves #206


lib/less/variables/colors.less
```diff
@@ -195,7 +195,7 @@
     error-color-hsl:            var(--red-400-h) var(--red-400-s) var(--red-400-l);
     warning-color:              var(--yellow-500);
     warning-color-hover:        var(--yellow-600);
-    warning-color-focus:        var(--yellow-200);
+    warning-color-focus:        var(--yellow-400);
```
ryandialpad, line 198: Is this change intended? I am guessing that the contrast ratio was not good enough

ryandialpad (approved):
w/ one question

Great work!

lib/less/variables/colors.less
dialpad-drew: Sure is and good guess! Explanation at the top of the PR.

ryandialpad: 🤦 Thanks!

PR #247, opened 2021-03-26
Svg utilities

dialpad-drew:
## Description
@hynes-dialpad would like to make sure you see this since this is finishing the work you'd started. 

This PR adds utility classes for SVG `fill` and documentation for SVG `size` and `fill`. 

**Note**: you'll see I'm using the existing icon size classes. They, unlike our other utility classes, use a double hyphen because BEM. I'm not entirely sure what the vision was here -- was it to move the "icon size" documentation to become SVG `size` documentation altogether? If so, then how to handle this is straight-forward. If not, then I'm not sure how to handle the naming.


ryandialpad (approved):
Looks good to me!

PR #270, opened 2021-04-07
Adds a page layout without a TOC

dialpad-drew:
## Description
This PR adds a page layout without a TOC. It also cleans up the doc site grid's `[content]` width so the pages are more consistently laid out. Previously, `min-content` could make some of the layout proportions different from page-to-page.

This PR also hides the side navigation on `md` breakpoints for the time-being. This way we can test the main content's and documentation's responsive.

**Note:** I don't really like the layout naming and would prefer a little more intention behind the name, but that's not where we are at the moment.


docs/assets/css/dialtone-docs.css
```diff
@@ -263,11 +263,20 @@ pre[class*="language-"].line-numbers > code {
   grid-area: toc !important;
 }
 .d-gl-docsite {
+  grid-column-gap: 4.8rem;
+          grid-column-gap: 4.8rem;
```
ryandialpad, line 267: Looks like this is duplicated?

ryandialpad: Looking below, it looks intentional. Do you know the reason for the duplication?

ryandialpad (approved):
w/ comment

Great Work!

docs/assets/css/dialtone-docs.css
dialpad-drew: Hmmm, I don't. This is a generated filed. Let me re-build and see if it persists.

dialpad-drew: I see it locally after re-builds.I think it's in the LESS-to-CSS compilation and may only be related to `column-gap`. I can replicate it on other selectors. I'll open an issue.

```
.toc li > ol {
  margin-left: var(--su16);
  column-gap: 4.8rem;
}
```
...compiles to...
```
.toc li > ol {
  grid-column-gap: 4.8rem;
        grid-column-gap: 4.8rem;
  margin-left: var(--su16);
  -webkit-column-gap: 4.8rem;
           -moz-column-gap: 4.8rem;
                column-gap: 4.8rem;
}
```

PR #272, opened 2021-04-08
V6 component tables docs

dialpad-drew:
## Description
PR does a couple things:
1. Updates documentation styles as in other documentation PRs
2. Removes background-color from `table`; I realized these inverted styles should behave like our other inverted styles, in which they rely on other elements to provide a background color.
3. Changes table striping to use `background-colors` with non-opaque alpha values rather than fully opaque colors; this way stripes can live on different colored light or dark backgrounds and still maintain their striping.
4. Added a table caption in examples, documentation, and class


lib/less/components/tables.less
```diff
@@ -91,7 +97,6 @@
     --table--th:           var(--black-050);
     --table--td:           var(--black-100);
     --table--bc:           var(--black-050);
-    --table--bgc:          var(--black-800);
```
ryandialpad, line 94: Are we removing cell background colors from the styles?

lib/less/components/tables.less
```diff
@@ -31,9 +30,16 @@
...
 
-    background-color: var(--table--bgc);
+    // Caption styles
+    .d-table__caption {
+      color: var(--table--th);
+      font-size: var(--fs18);
+      font-weight: var(--fw-bold);
+      margin-bottom: var(--su8);
+      .d-ta-unset();
+    }
```
ryandialpad, line 42: Did you want to select the captions using the element instead of a classname? This will prevent a situation where a user forgets to add the caption class to their caption.

ryandialpad (approved):
w/ A comment and a question

Great work!

lib/less/components/tables.less
dialpad-drew: Yea, I think we should. See comment in PR. 👆

lib/less/components/tables.less
dialpad-drew: I'm not sure and am on the fence about. That's how I originally did it, but then thought there are likely instances in which captions may need other styles. Sometimes a big header isn't appropriate, and figured styling the tag rather than a class means those styles have higher priority in the cascade. It's probably not a big deal if a dev uses our utility classes and `!important` overrides that, though.

I don't think I feel strongly about it one way or the other. What do you think?

ryandialpad: If there are cases where other styles are needed then I would leave this as it's own class

PR #276, opened 2021-04-13
V6 component toast docs

dialpad-drew:
## Description
Updates Toast documentation.
1. Cleans up some of the class table organization and JSON file
2. Adds A11y blurb, though we'll need to re-visit to mimic Vue a11y implementation
3. Removes positioning class and bakes it into the component
4. Adds code to `codeWellFooter`

**Note**: I'll need to revisit this component in a future PR to add additional use cases.


ryandialpad (approved):
Great work!

PR #280, opened 2021-04-13
add new icon MS teams device

raodialpad:
## Description
To add a new icon MS Teams Device, which will be used to represent a type of device in dialpad web settings -> Your user devices (replace '#' in the red rectangle).
image

The new icon is:
image

## Pull Request Checklist


dialpad-drew:
hey @raodialpad , thanks for the PR. Are you working with a designer on this? I'm curious where this icon came from.

If you're not working with a designer, is there a reason we couldn't use `<IconMicrosoftTeams />` by itself rather than combining it with `<IconHashtag />`?

raodialpad:
> hey @raodialpad , thanks for the PR. Are you working with a designer on this? I'm curious where this icon came from.
> 
> If you're not working with a designer, is there a reason we couldn't use `<IconMicrosoftTeams />` by itself rather than combining it with `<IconHashtag />`?

Thanks Drew for the review, the new icon is created by Vicky @vcheng-dialpad, I think the reason is that we use it as a type of device and to differentiate it from the MS Teams brand icon that will be used in other scenarios.

dialpad-drew:
> Thanks Drew for the review, the new icon is created by Vicky @vcheng-dialpad, I think the reason is that we use it as a type of device and to differentiate it from the MS Teams brand icon that will be used in other scenarios.

I see. So, since this table is representing a list of devices and it sounds like this icon represents only a single device on which MS Teams is installed, a user would have to add each individual device on which they have MS Teams (i.e. their laptop, their phone, their tablet, etc.). Am I understanding it correctly?

raodialpad:
> I see. So, since this table is representing a list of devices and it sounds like this icon represents only a single device on which MS Teams is installed, a user would have to add each individual device on which they have MS Teams (i.e. their laptop, their phone, their tablet, etc.). Am I understanding it correctly?

We do not tell the platform differences for Teams, we treat them all as one single device, so, for example, if user installed Teams on both laptop and phone, our our site, it is just one MS Teams device. The MS Teams device here is not an real device, it is an endpoint.

dialpad-drew:
> > I see. So, since this table is representing a list of devices and it sounds like this icon represents only a single device on which MS Teams is installed, a user would have to add each individual device on which they have MS Teams (i.e. their laptop, their phone, their tablet, etc.). Am I understanding it correctly?
> 
> We do not tell the platform differences for Teams, we treat them all as one single device, so, for example, if user installed Teams on both laptop and phone, our our site, it is just one MS Teams device.

That makes much more sense and has been my experience as a previous MS Teams users. Anyone calling me on Teams calls me on all devices, not just a single device, which is the crux of my question to Vicky as to why we wouldn't just show the MS Teams app icon. 

Thanks for the clarification. Talking to Vicky now -- will get back to you soon!

docs/_data/svg-system.json
```diff
@@ -607,6 +607,12 @@
     "vue": "IconMoments",
     "desc": "Communicates that an item is a Dialpad Vi Moment."
   },
+  {
+    "name": "MS Teams Device",
+    "file": "ms-teams-device",
+    "vue": "MSTeamsDevice",
```
raodialpad: I think this should be 'MsTeamsDevice', will make the change.

raodialpad:
closing this PR cause we will use current MS Teams icon.

PR #283, opened 2021-04-14
More doc clean-up

dialpad-drew:
## Description
This PR does a few very small things but with many changes:
* Updates site nav to go to each sections' first page instead of a category page. For instance, clicking on Utilities now takes users to the first utility rather than an empty page. The category pages would need to be designed and developed, and there's just not time at the moment. It's a nice-to-have.
* Updates components' class table styles to be more consistently styled -- they differed from component-to-component. Descriptions are also no longer monospace
* Replaces the term "Default" with "Base Style(s)" when referring to style. The term "Default" should be reserved when discussing state (i.e. default, focus, hover, active, etc.). Figma has already been updated.
* Changes components' to be singular except for those components that are inherently plural (i.e. Breadcrumbs, Tabs)
  * Changes JSON file names
  * Changes HTML file names
  * Changes documentation JS file names
  * Changes LESS file names
  * Changes LESS imports


ryandialpad (approved):
LGTM!

I see many merge conflicts in my future 🤯

PR #289, opened 2021-04-16
Clean up Adding icon and favicon documentation

dialpad-drew:
## Description
This PR cleans up "Adding icon" documentation and cleans up and removes the "Favicon" documentation, which needs more work than we have time at the moment.

### DIFF
--- a/docs/design/icons/adding.html
+++ b/docs/design/icons/adding.html
@@ -1,35 +1,37 @@
 ---
-layout: page
+layout: page-no-toc
 title: Adding Icons
 description: A quick guide for how to add icons to Dialtone.
 ---
-<section class="d-layout__stack16">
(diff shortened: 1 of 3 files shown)


docs/design/icons/adding.html
```diff
@@ -1,35 +1,37 @@
 ---
```
ryandialpad, line 1: Nit: Would it be possible to shorten some of the line lengths here, just to make future editing easier.

docs/design/icons/favicons.html
```diff
@@ -4,40 +4,36 @@
 description: Dialtone provides various favicons that can be used by our various products and states.
```
ryandialpad, line 4: Same comment as above

ryandialpad (approved):
w/ two minor comments

Issue #253, opened 2021-03-29
[CSS] Component: Select Menus


ryandialpad:
**Acceptance Criteria**

- Update the Select Menu component Dialtone classes

**Resources**

Design: https://www.figma.com/file/zP2uyVxAg4notGNZwW64mW/Components-(Dialtone-6.0)?node-id=141%3A0

PR #291, opened 2021-04-19
V6 [CSS] Component: Select Menus

nikitadialpad:
## Description
- #253 [CSS] Component: Select Menus. Update the Select Menu component DIaltone classes
- `v6-components-select` was used as base branch due to the clarification

## Pull Request Checklist

## Obligatory GIF
.gif)

### DIFF
--- a/lib/less/components/selects.less
+++ b/lib/less/components/selects.less
@@ -12,7 +12,7 @@
 //
 
 .d-select-disable {
-    --select--fc: #151526;
+    --select--fc: #151526; // https://htmlcolors.com/hex-to-hsl
 
     background: #CBCBD9;
     border: 1px solid #47475C;
@@ -34,9 +34,8 @@
   font-size: var(--fs16);
   font-family: inherit;
   line-height: var(--lh-tight);
-  margin-top: 6px;
 
-  //  --  Creat some arrows
+  //  --  Arrows
   &:before,
   &:after {
       content: "";
@@ -112,7 +111,7 @@
   .input-button-lg();
   font-size: var(--fs20);
   padding-left: .7rem;
-  padding-right: 2em;
+  padding-right: 2em; // TODO Select CSS Vars
 }
 .d-select--xl > select {
   .input-button-xl();
@@ -193,7 +192,7 @@
 //  ============================================================================
 //  $  LABEL FOR SELECT (NATIVE)
 //  ----------------------------------------------------------------------------
-.d-select__label {
+.d-label label {
     .d-label();
     color: #151526;
     margin-bottom: 0;
@@ -204,24 +203,15 @@
     }
 }
 
-.d-select__label--xs {
-    .d-label--xs()
-}
-.d-select__label--sm {
-    .d-label--sm()
-}
-.d-select__label--lg {
-    .d-label--lg();
-
-    ~ .d-description--lg {
-        line-height: var(--lh4);
-    }
-}
-.d-select__label--xl {
-    .d-label--xl();
-
-    ~ .d-description--xl {
-        font-size: var(--fs16);
-    }
-}
-
+// .d-select__label--xs {
+//     .d-label--xs()
+// }
+// .d-select__label--sm {
+//     .d-label--sm()
+// }
+// .d-select__label--lg {
+//     .d-label--lg();
+// }
+// .d-select__label--xl {
+//     .d-label--xl();
+// }
--- a/lib/less/components/selects.less
+++ b/lib/less/components/selects.less
@@ -197,30 +197,31 @@
     .d-label();
     color: #151526;
     margin-bottom: 0;
+
+    ~ .d-description {
+        margin-bottom: 0;
+        color: #47475C;
+    }
 }
 
 .d-select__label--xs {
-    font-size: var(--fs12);
+    .d-label--xs()
 }
 .d-select__label--sm {
-    font-size: var(--fs14);
+    .d-label--sm()
 }
 .d-select__label--lg {
-    font-size: var(--fs20);
+    .d-label--lg();
+
+    ~ .d-description--lg {
+        line-height: var(--lh4);
+    }
 }
 .d-select__label--xl {
-    font-size: var(--fs24);
+    .d-label--xl();
+
+    ~ .d-description--xl {
+        font-size: var(--fs16);
+    }
 }
 
-.d-select__description {
-    .d-description();
-    margin-bottom: 0;
-    color: #47475C;
-}
-.d-select__description--lg {
-    .d-description--lg();
-    line-height: var(--lh4);
-}
-.d-select__description--xl {
-    font-size: var(--fs16);
-}
--- a/lib/less/components/selects.less
+++ b/lib/less/components/selects.less
@@ -11,7 +11,7 @@
 //    - Sizes
 //
 
-.d-select-disable {
+.d-select--disable {
     background: var(--black-200);
     border: 1px solid var(--black-500);
 }
@@ -76,7 +76,7 @@
       }
 
       &:disabled {
-        .d-select-disable()
+        .d-select--disable();
       }
 
       &::-moz-focus-inner {
@@ -126,7 +126,7 @@
         }
 
         &:disabled {
-            .d-select-disable()
+            .d-select--disable()
         }
     }
 
@@ -148,7 +148,7 @@
         }
 
         &:disabled {
-            .d-select-disable()
+            .d-select--disable()
         }
     }
 
@@ -171,7 +171,7 @@
         }
 
         &:disabled {
-            .d-select-disable()
+            .d-select--disable()
         }
     }
 
--- a/lib/less/components/selects.less
+++ b/lib/less/components/selects.less
@@ -11,7 +11,7 @@
 //    - Sizes
 //
 
-.d-select--disable {
+.d-select--disable() {
     background: var(--black-200);
     border: 1px solid var(--black-500);
 }
--- a/lib/less/components/selects.less
+++ b/lib/less/components/selects.less
@@ -50,7 +50,7 @@
       border-bottom-width: 0;
   }
 
-  > select {
+  .d-select__input {
       //  [1] Reset the appearance
       -webkit-appearance: none;
          -moz-appearance: none;
@@ -92,29 +92,40 @@
 //  $$  SIZES
 //  ----------------------------------------------------------------------------
 
-.d-select--xs > select {
-  .input-button-xs();
-  padding-right: var(--su24);
-}
-.d-select--sm > select {
-  .input-button-sm();
-  padding-right: var(--su24);
-}
-.d-select--lg > select {
-  .input-button-lg();
-  padding-left: var(--su8);
-  padding-right: var(--su24);
-}
-.d-select--xl > select {
-  .input-button-xl();
-  padding-left: var(--su8);
-  padding-right: var(--su24);
-}
+.d-select--xs {
+  .d-select__input {
+    .input-button-xs();
 
+    padding-right: var(--su24);
+  }
+}
+.d-select--sm {
+  .d-select__input {
+    .input-button-sm();
+
+    padding-right: var(--su24);
+  }
+}
+.d-select--lg {
+  .d-select__input {
+    .input-button-lg();
+
+    padding-left: var(--su8);
+    padding-right: var(--su24);
+  }
+}
+.d-select--xl {
+  .d-select__input {
+    .input-button-xl();
+
+    padding-left: var(--su8);
+    padding-right: var(--su24);
+  }
+}
 //  $$  VALIDATION STATES
 //  ----------------------------------------------------------------------------
 .d-select--warning {
-    > select {
+    .d-select__input {
         --select--bc: var(--yellow-500);
 
         &:focus {
@@ -132,7 +143,7 @@
     }
 }
 .d-select--error {
-    > select {
+    .d-select__input {
         --select--bc: var(--red-400);
 
         &:focus {
@@ -151,7 +162,7 @@
 }
 
 .d-select--success {
-    > select {
+    .d-select__input {
         --select--bc: var(--green-400);
 
         &:focus {
--- a/lib/less/components/selects.less
+++ b/lib/less/components/selects.less
@@ -91,7 +91,6 @@
 
 //  $$  SIZES
 //  ----------------------------------------------------------------------------
-
 .d-select--xs {
   .d-select__input {
     .input-button-xs();
@@ -144,7 +143,6 @@
         }
     }
 }
-
 .d-select--success {
     .d-select__input {
         --select--bc: var(--green-400);
--- a/lib/less/components/selects.less
+++ b/lib/less/components/selects.less
@@ -10,6 +10,13 @@
 //  • SELECT MENUS
 //    - Sizes
 //
+
+.d-select-disable {
+    --select--fc: #151526;
+
+    background: #CBCBD9;
+    border: 1px solid #47475C;
+}
 //  ============================================================================
 //  $  SELECT MENU (NATIVE)
 //  ----------------------------------------------------------------------------
@@ -27,6 +34,7 @@
   font-size: var(--fs16);
   font-family: inherit;
   line-height: var(--lh-tight);
+  margin-top: 6px;
 
   //  --  Creat some arrows
   &:before,
@@ -57,12 +65,23 @@
 
       //  [2] Update the styles
       .d-input();
+      --select--bc: #47475C;
+      --select--fc: #151526;
       padding: 0.7rem 2em 0.7rem 0.7rem;
       height: 100%; // Fill the height of its parent
 
       color:            var(--select--fc);
       background-color: var(--select--bgc);
       border-color:     var(--select--bc);
+
+      &:focus {
+        --select--bc: #1635E5;
+        box-shadow: 0 0 0 4px #89b8ff;
+      }
+
+      &:disabled {
+        .d-select-disable()
+      }
 
       &::-moz-focus-inner {
           outline: none !important;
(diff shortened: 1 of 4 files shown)


lib/less/components/selects.less
```diff
@@ -81,44 +100,127 @@
...
 }
 .d-select--lg > select {
   .input-button-lg();
+  font-size: var(--fs20);
+  padding-left: .7rem;
   padding-right: 2em;
 }
 .d-select--xl > select {
   .input-button-xl();
+  font-size: var(--fs24);
```
dialpad-drew: Our font-sizes are close but off. 

`xs`: correct
`sm`: correct
`md`: 16px
`lg`: 18px
`xl`: 20px

lib/less/components/selects.less
```diff
@@ -81,44 +100,127 @@
...
+}
+.d-select__label--sm {
+    font-size: var(--fs14);
+}
+.d-select__label--lg {
+    font-size: var(--fs20);
+}
+.d-select__label--xl {
+    font-size: var(--fs24);
+}
```
dialpad-drew: Our font-sizes are close but off. 

`xs`: correct
`sm`: correct
`md`: 16px
`lg`: 18px
`xl`: 20px

dialpad-drew (commented):
The mark-up is close, but not quite right. `.d-select` should be on the `select` element, not wrapping it. The CSS targeting `.d-select > select` would need to be modified accordingly.

CC: @ryandialpad because he did a lot of work on inputs.

lib/less/components/selects.less
ryandialpad: Font sizes are being redefined. The font sizes are already defined in `.input-button--<SIZE>`

See the following file: https://github.com/dialpad/dialtone/blob/v6-components-inputs/lib/less/variables/sizes.less

lib/less/components/selects.less
ryandialpad: You can use the label classes defined in the following file: https://github.com/dialpad/dialtone/blob/v6-components-inputs/lib/less/components/forms.less

lib/less/components/selects.less
```diff
@@ -81,44 +100,127 @@
...
+
+.d-select__description {
```
ryandialpad: Please use the description classes defined in the following file: https://github.com/dialpad/dialtone/blob/v6-components-inputs/lib/less/components/forms.less

lib/less/components/selects.less
```diff
@@ -81,44 +100,127 @@
...
+    .d-description();
+    margin-bottom: 0;
+    color: #47475C;
```
ryandialpad: Please use existing named colours.

The figma design panel should have a named colour swatch which will be defined in the following file: https://github.com/dialpad/dialtone/blob/v6-components-inputs/lib/less/variables/colors.less

Example Usage

```css
.my-class {
  color: var(--my-named-colour);
}
```

ryandialpad:
> The mark-up is close, but not quite right. `.d-select` should be on the `select` element, not wrapping it. The CSS targeting `.d-select > select` would need to be modified accordingly.
> 
> CC: @ryandialpad because he did a lot of work on inputs.

@dialpad-drew Agreed 👍 

The inputs and textarea classes are applied onto their respective elements.

Examples:

Inputs
```html
<label class="d-label" for="my-input">...</label>
<input class="d-input d-input--<VARIANT>" id="my-input" type="email" placeholder="..." value="..." />
<span class="d-validation-message d-validation-message--<VARIANT>">...</span>
```

Textareas
```html
<label class="d-label" for="my-textarea">...</label>
<textarea class="d-textarea d-textarea--<VARIANT>" id="my-textarea" type="email" value="...">...</textarea>
<span class="d-validation-message d-validation-message--<VARIANT>">...</span>
```

Selects
```html
<label class="d-label" for="my-select">...</label>
<select class="d-select d-select--<VARIANT>" id="my-select">
    <option>…</option>
    ...
</select>
<span class="d-validation-message d-validation-message--<VARIANT>">...</span>
```

lib/less/components/selects.less
nikitadialpad: ok, sure, pushed the changes according the the comment.

But need to ask, just to double check this case - according to the design (attached screenshot) the font sizes are a bit different 
Screenshot 2021-04-22 at 13 06 06

lib/less/components/selects.less
nikitadialpad: ok, sure, pushed the changes according the the comment.

The comment resolution is depend on answer for previous comment.

lib/less/components/selects.less
nikitadialpad: changed, used `~` for styles aligning

this will be aligned as well due to last comments

lib/less/components/selects.less
```diff
@@ -81,44 +100,128 @@
...
+        color: #47475C;
+    }
+}
+
+.d-select__label--xs {
+    .d-label--xs()
 }
+.d-select__label--sm {
+    .d-label--sm()
+}
```
ryandialpad: Can we use the generic `d-label` classes instead of using them in mixins here?

lib/less/components/selects.less
```diff
@@ -81,44 +100,128 @@
...
+.d-select__label--lg {
+    .d-label--lg();
+
+    ~ .d-description--lg {
+        line-height: var(--lh4);
+    }
```
ryandialpad: Can we avoid nesting the description classes within the label classes?

@dialpad-drew Is there a reason why the design for the description large variant differs for the large select menu variants?

lib/less/components/selects.less
```diff
@@ -81,44 +100,128 @@
...
+}
+.d-select__label--xl {
+    .d-label--xl();
+
+    ~ .d-description--xl {
+        font-size: var(--fs16);
+    }
```
ryandialpad: @dialpad-drew See above comment

lib/less/components/selects.less
ryandialpad: The font sizes in the design line up with the font sizes defined in the sizes file and base input class


Copied from `sizes.less`

```css
//  $   SHARED INPUT & BUTTON SIZES
//  $$  EXTRA SMALL
//      Target height: 24px
.input-button-xs() {
    padding-top:    var(--su6);
    padding-bottom: var(--su6);
    font-size:      var(--fs12);
    ....
}

//  $$  SMALL
//      Target height: 32px
.input-button-sm() {
    padding-top:    0.7rem;
    padding-bottom: 0.7rem;
    font-size:      var(--fs14);
    ...
}

//  $$  LARGE
//      Target height: 48px
.input-button-lg() {
    padding-top:    var(--su12);
    padding-bottom: var(--su12);
    font-size:      var(--fs20);
    ...
}

//  $$  EXTRA LARGE
//      Target height: 56px
.input-button-xl() {
    padding-top:    1.4rem;
    padding-bottom: 1.4rem;
    font-size:      var(--fs24);
    ...
}
```

The medium variant is defined in `select.less` as it is the base select style

ryandialpad: @dialpad-drew Are we changing the sizes for selects from the already defined sizes for inputs and textareas?

It looks like the sizes you mentioned above do not line up with the current design / size variants that were defined for inputs and textareas.

lib/less/components/selects.less
ryandialpad: Usage from an `inputs` example:

```html
<label class="d-label" for="Dialtone--InputExample--ExtraSmall">Full name</label>
<span class="d-description">This is something you can change later.</span>
<span class="d-input-icon d-input-icon--xs d-input-icon--left">{% iconSystem "send" %}</span>
<span class="d-input-icon d-input-icon--xs d-input-icon--right">{% iconSystem "lock" %}</span>
<input class="d-input d-input--xs" id="Dialtone--InputExample--ExtraSmall" type="text" placeholder="e.g. John Doe, Jane Doe" />
```

ryandialpad: One note with this example is that there is on accessibility concern and that is the description should describe the input. Screen readers would probably have an issue with the current markdown.

lib/less/components/selects.less
ryandialpad: > > The mark-up is close, but not quite right. `.d-select` should be on the `select` element, not wrapping it. The CSS targeting `.d-select > select` would need to be modified accordingly.
> > CC: @ryandialpad because he did a lot of work on inputs.
> 
> @dialpad-drew Agreed 👍
> 
> The inputs and textarea classes are applied onto their respective elements.
> 
> Examples:
> 
> Inputs
> 
> ```
> <label class="d-label" for="my-input">...</label>
> <input class="d-input d-input--<VARIANT>" id="my-input" type="email" placeholder="..." value="..." />
> <span class="d-validation-message d-validation-message--<VARIANT>">...</span>
> ```
> 
> Textareas
> 
> ```
> <label class="d-label" for="my-textarea">...</label>
> <textarea class="d-textarea d-textarea--<VARIANT>" id="my-textarea" type="email" value="...">...</textarea>
> <span class="d-validation-message d-validation-message--<VARIANT>">...</span>
> ```
> 
> Selects
> 
> ```
> <label class="d-label" for="my-select">...</label>
> <select class="d-select d-select--<VARIANT>" id="my-select">
>     <option>…</option>
>     ...
> </select>
> <span class="d-validation-message d-validation-message--<VARIANT>">...</span>
> ```

@dialpad-drew I did some investigation and we will need a wrapper if we want to keep our custom arrows on the select element.

Let me know what you think of the following usage examples

1.
```html
<div class="d-select__wrapper">
  <select class="d-select">
    ...
  </select>
</div>
```

2.
```html
<div class="d-label">...</div>
<div class="d-description">...</div>
<div class="d-select">
  <select class="d-select__input">
    ...
  </select>
</div>
  <div class="d-validation-message">...</div>
```

3.
```html
<div class="d-select">
  <div class="d-label">...</div>
  <div class="d-description">...</div>
  <select class="d-select__input">
    ...
  </select>
  <div class="d-validation-message">...</div>
</div>
```

In terms of my preference is either 1 or 2

cc @nikhildialpad

lib/less/components/selects.less
dialpad-drew: In which aspect(s)? I'd prefer not to dig through each of these variants trying to uncover the reference.

lib/less/components/selects.less
dialpad-drew: No. Just use the sizes from the other inputs. Something is wonky between here and Figma; I'll come back through and fix it later.

lib/less/components/selects.less
ryandialpad: Here is the current definitions for our form description elements

```css
//  $ DESCRIPTIONS
.d-description {
  box-sizing: border-box;

  display: flex;
  margin: var(--su0);
  color: var(--black-500);
  fill: currentColor;
  font-size: var(--fs12);
  font-family: inherit;
  line-height: var(--lh4);
}

//  $$  SIZES
.d-description--lg {
  font-size:   var(--fs14);
  line-height: var(--lh6);
}
```

We are missing an XL variant and the line height differs between the large description used in inputs and textareas and the large description defined in selects.

Did we want to use a single set of description classes for all of our forms? If so, did you want to update the existing large description variant to have a line height of `lh4` instead of `lh6`?

lib/less/components/selects.less
dialpad-drew: I think the first is reasonable, though do you think the second would be more backwards compatible? Not really into the third.

ryandialpad: @dialpad-drew Wasn't a fan of the third either, just added it to show all of the options that I could think of.

The second would not be more backwards compatible as we would ideally move the select styles into the class that is applied to the select element

lib/less/components/selects.less
nikitadialpad: ok, got it, tnx

lib/less/components/selects.less
ryandialpad: We don't currently have any `__wrapper` classes defined in dialtone but we do have `__input` classes defined. I think we should lean towards the already established patterns unless you would prefer otherwise.

dialpad-drew: Agreed.

lib/less/components/selects.less
dialpad-drew: Use the existing styles. Like I mentioned in another comment, there's something wonky going on, so after all this, I'm going to go back through Figma and make sure things are tight, and come back through the CSS and do the same.

dialpad-drew: Also, thank you for clarifying. :)

lib/less/components/selects.less
nikitadialpad: ok, got it, tnx

lib/less/components/selects.less
ryandialpad: Anytime :)

dialpad-drew:
Hey @nikitadialpad , looks like you're cranking. 🚀  Let @ryandialpad and I know when you're ready for another review. 🙌

lib/less/components/selects.less
nikitadialpad: done, according to the latest changes

lib/less/components/selects.less
nikitadialpad: done

lib/less/components/selects.less
nikitadialpad: done, according to the latest changes, 
used default classes and styles as was discussed previously

lib/less/components/selects.less
nikitadialpad: done, according to the latest changes,
used default classes and styles as was discussed previously

lib/less/components/selects.less
nikitadialpad: done

lib/less/components/selects.less
```diff
@@ -10,6 +10,11 @@
 //  • SELECT MENUS
 //    - Sizes
 //
+
+.d-select-disable {
```
ryandialpad:
```suggestion
.d-select--disable {
```

We typically use `--` to break up variants

ryandialpad: Do we need to define this separately if we are using it below as a mixin on the select class when the disabled state is detected?

This classes should be chained so `d-select` will always be present

lib/less/components/selects.less
ryandialpad: Leaving unresolved so I can refer to this conversation during review

nikitadialpad:
hey @dialpad-drew, could you please paraphrase the first part of the last sentence, sorry but I didn't get it.

lib/less/components/selects.less
nikitadialpad: sure, 
`d-select--disable` was switched to mixing
are you ok if keep this mixing on the top of select.less file? in other case could you please mention where it could be moved

ryandialpad: If the mixin is only being used in a single place `&:disabled` it might be better to moved the styles into the `&:disabled` block.

The component should support class chaining so `d-select` and `d-select__input` should always be defined

nikitadialpad: it was used in several cases, due to design purpose, 
after last comments, it's not necessary 

so removed the mixing

lib/less/components/selects.less
```diff
@@ -57,13 +58,24 @@
```
ryandialpad, line 61: Did we want to move `> select` styles into the `d-select__input` class?

We can also update the variant classes such as `d-select--<SIZE>` and `d-select--<STATE>` to select on our `d-select__input` class

nikitadialpad: hm, I'm out of the discussion. as for me.. why don't have one class in this case?
I mean may be` d-select__input` which sounds more semantic / consistent with `d-input` class. 

I mean this

```
<div class="d-select__input">
  <select>
    ...
  </select>
</div>
```

ryandialpad: This is in reference to the discussion here: 

https://github.com/dialpad/dialtone/pull/291#discussion_r618433558
https://github.com/dialpad/dialtone/pull/291#discussion_r618508125

nikitadialpad: hm, according to those discussions the last question was asked.
ok, never mind, the styles were changed

nikitadialpad: done

lib/less/components/selects.less
```diff
@@ -79,46 +91,90 @@
...
+            --select--bc: var(--green-400);
+            box-shadow: 0 0 0 4px var(--green-200);
+        }
+    }
+
+    & ~ .d-validation-message--success {
+        --success-color: var(--green-500);
+        &::before {
+            background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' viewBox='0 0 24 24' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm-2 15l-5-5 1.41-1.41L10 14.17l7.59-7.59L19 8l-9 9z' fill='hsla(162, 87%, 21%, 1)'/%3E%3C/svg%3E");
+        }
```
ryandialpad: It looks like we are re-defining the validation message styles which are defined in `forms.less`

lib/less/components/selects.less
```diff
@@ -79,46 +91,90 @@
...
+        --warning-color: var(--yellow-500);
 
-    &:focus {
-        box-shadow: 0 0 0 var(--su1) var(--white),
-                    0 0 0.25em var(--warning-color-focus);
+        &::before {
+            background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' viewBox='0 0 24 24' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M1 21h22L12 2 1 21zm12-3h-2v-2h2v2zm0-4h-2v-4h2v4z' fill='hsla(40, 100%, 31%, 1)'/%3E%3C/svg%3E");
+        }
     }
 }
```
ryandialpad: See below

lib/less/components/selects.less
```diff
@@ -79,46 +91,90 @@
...
+        }
+    }
+
+
+    & ~ .d-validation-message--error {
+        --error-color: var(--red-400);
+        &::before {
+            background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' viewBox='0 0 24 24' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm-2 15l-5-5 1.41-1.41L10 14.17l7.59-7.59L19 8l-9 9z' fill='hsla(356, 79%, 51%, 1)'/%3E%3C/svg%3E");
+        }
     }
```
ryandialpad: See below

nikitadialpad: removed

lib/less/components/selects.less
nikitadialpad: removed

lib/less/components/selects.less
nikitadialpad: removed

lib/less/components/selects.less
```diff
@@ -79,46 +91,65 @@
...
+        &:focus {
+            --select--bc: var(--red-400);
+            box-shadow: 0 0 0 4px var(--red-200);
+        }
     }
 }
-.d-select--success > select {
-    --select--bc:       var(--success-color);
+.d-select--success {
+    .d-select__input {
```
ryandialpad: Final set of comments, we can now merge these two classes like so `.d-select-success.d-seelect__input {}`. There are two other occurrences of this.

ryandialpad (approved):
w/ one minor comment for a few revisions

Pending @dialpad-drew 's approval

Great work 🚀

lib/less/components/selects.less
ryandialpad: Based on the conversation we had on slack, I think that the better approach that more closely aligns with BEM as you pointed out would be to rename these classes to `d-select__input--<validation_state>`

dialpad-drew:
> hey @dialpad-drew, could you please paraphrase the first part of the last sentence, sorry but I didn't get it.

Oh, sorry. I just meant to say it looked like you were doing a lot of work. 👍

lib/less/components/selects.less
```diff
@@ -49,21 +50,32 @@
...
       .d-input();
+      --select--bc: var(--black-500);
       padding: 0.7rem 2em 0.7rem 0.7rem;
       height: 100%; // Fill the height of its parent
 
       color:            var(--select--fc);
       background-color: var(--select--bgc);
       border-color:     var(--select--bc);
 
+      &:focus {
```
dialpad-drew: Focus doesn't have the space between the input and the focus-ring.

dialpad-drew: Example: 
Screen Shot 2021-04-27 at 10 21 13 AM

dialpad-drew (approved):
Nice job. With the exception of Ryan's feedback and my focus-ring spacing feedback, this looks good to me.

lib/less/components/selects.less
nikitadialpad: hm, according to the design in the section at the bottom

Screenshot 2021-04-27 at 19 41 06

do we need this space, right?

dialpad-drew: Ahh, you're right. That's a bug in the design file. Sorry about that.

Yes, we want the space. Ryan did it with Inputs, Checkboxes, and Radios, so there should be an existing pattern there if you're curious how they did it.

nikitadialpad: ok, got it, thank you, will do

nikitadialpad:
> > hey @dialpad-drew, could you please paraphrase the first part of the last sentence, sorry but I didn't get it.
> 
> Oh, sorry. I just meant to say it looked like you were doing a lot of work. 👍

thank you :)

lib/less/components/selects.less
nikitadialpad: done

nikitadialpad:
@ryandialpad could I ask you for review again, and If you don't mind could you please check this branch and test it locally to avoid some design mistakes, because my vision is a bit blurry after those changes :)

lib/less/components/selects.less
```diff
@@ -24,101 +25,129 @@
...
+      display: none;
+    }
+    //  $$  VALIDATION STATES
+    //  ------------------------------------------------------------------------
+    &--success {
+      --select--bc: var(--green-400);
 
-    &:focus {
+      &:focus {
+        --select--bc: var(--green-400);
```
ryandialpad, line 97: One nit pick to clean this up a bit is that I would define a focus colour variable that can be changed at the top level of the variant like so:

```css
.d-select__input--warning {
  --select-bc: var(--yellow-500);
  --select--focus-bc:  var(--yellow-500);
}
```

It keeps the code a little cleaner and reduces the potential risk of a side effect by changing the select background colours multiple times

ryandialpad (approved):
w/ one nit pick should you choose to address it

Great Work 🚀 

@dialpad-drew We should look into creating a focus ring utility in the future to DRY up some of this code. Maybe in DT 6.1?

lib/less/components/selects.less
nikitadialpad: done

ryandialpad: Sorry for the confusion, can we revert this commit for now and look into a more holistic approach in the future. The double nesting of vars may potentially cause issues in browsers.

ryandialpad: After the revert feel free to merge

nikitadialpad: reverted, done

nikitadialpad: @ryandialpad 
`Only those with write access to this repository can merge pull requests.`
I'm not able to merge now

PR #293, opened 2021-04-23
V6 Component: Inputs documentation

dialpad-drew:
## Description
This PR:
* Combines Input and Textarea documentation into a single Input page because they're a single Vue component and a single Figma component; documentation should reflect that architecture
* Adds A11y documentation to Input
* Removes `.has-an-icon--left` and `.has-an-icon--right` classes
* Removes Forms page for now until we can better and holistically organized form component documentation
* Removes `textareas.json`


dialpad-drew:
@ryandialpad Do the `textareas'` validation messages look wonky to you? They seem pretty disconnected from the `input` to me.

ryandialpad:
> @ryandialpad Do the `textareas'` validation messages look wonky to you? They seem pretty disconnected from the `input` to me.

I did notice that, I inspected the styles and didn't notice and additional margins or padding. I am wondering if it is a browser style that is causing the element below the textarea being spaced out a bit more than expected.

docs/components/inputs.html
```diff
@@ -1,315 +1,419 @@
...
-<input class="d-input" id="Dialtone--InputExample1" type="text" placeholder="…" />
-{% endhighlight %}
-        </div>
-    </div>
+          {% endcodeWellHeader %}
+          {% codeWellFooter %}
+            {% highlight html linenos %}
+<div>
+  <label class="d-label" for="Dialtone--InputExample1a">...</label>
+  <span class="d-description">...</span>
```
ryandialpad:
```suggestion
```

docs/components/inputs.html
```diff
@@ -1,315 +1,419 @@
...
-                </span>
-            </div>
-            <div class="dialtone-example__code">
-{% highlight html linenos %}
 <div>
-    <label class="d-label" for="my-input">...</label>
-    <input class="d-input d-input--warning" id="my-input" type="email" placeholder="" value="some value" />
-    <span class="d-validation-message d-validation-message--warning">...</span>
+  <label class="d-label" for="Dialtone--InputExample1b">...</label>
+  <span class="d-description">...</span>
```
ryandialpad:
```suggestion
```

docs/components/inputs.html
ryandialpad: Remove description from code example since the example above does not include a description

docs/components/inputs.html
ryandialpad: Remove description from code example since the example above does not include a description

docs/components/inputs.html
```diff
@@ -1,315 +1,419 @@
...
-{% highlight html linenos %}
+          {% endcodeWellHeader %}
+          {% codeWellFooter %}
+            {% highlight html linenos %}
 <div>
-    <label class="d-label" for="my-input">...</label>
-    <input class="d-input d-input--error" id="my-input" type="email" placeholder="" value="some value" />
-    <span class="d-validation-message d-validation-message--error">...</span>
+  <label class="d-label" for="Dialtone--TextareaExample1a">...</label>
+  <span class="d-description">...</span>
```
ryandialpad:
```suggestion
```

ryandialpad: Remove description from code example since the example above does not include a description

docs/components/inputs.html
```diff
@@ -1,315 +1,419 @@
...
+  <textarea class="d-textarea" id="Dialtone--TextareaExample1a" type="text" placeholder="..."></textarea>
 </div>
-{% endhighlight %}
+<div>
+  <label class="d-label" for="Dialtone--TextareaExample1b">...</label>
+  <span class="d-description">...</span>
```
ryandialpad:
```suggestion
```

ryandialpad: Remove description from code example since the example above does not include a description

ryandialpad:
> > @ryandialpad Do the `textareas'` validation messages look wonky to you? They seem pretty disconnected from the `input` to me.
> 
> I did notice that, I inspected the styles and didn't notice and additional margins or padding. I am wondering if it is a browser style that is causing the element below the textarea being spaced out a bit more than expected.

Looks like my hunch was correct: https://stackoverflow.com/questions/3453959/how-do-i-fix-inconsistent-textarea-bottom-margin-in-firefox-and-chrome

We need to add `display: block;` to the textarea. I will add that to the class now.

The source of the issue is by default textareas will use text height which creates this extra margin at the bottom that does not show up when inspecting.

ryandialpad:
> > > @ryandialpad Do the `textareas'` validation messages look wonky to you? They seem pretty disconnected from the `input` to me.
> > 
> > 
> > I did notice that, I inspected the styles and didn't notice and additional margins or padding. I am wondering if it is a browser style that is causing the element below the textarea being spaced out a bit more than expected.
> 
> Looks like my hunch was correct: https://stackoverflow.com/questions/3453959/how-do-i-fix-inconsistent-textarea-bottom-margin-in-firefox-and-chrome
> 
> We need to add `display: block;` to the textarea. I will add that to the class now.
> 
> The source of the issue is by default textareas will use text height which creates this extra margin at the bottom that does not show up when inspecting.

Going to need to do some further investigation as adding `display: block;` causes issues with icon positioning in some cases.

ryandialpad:
Found a better solution here that also includes a good explanation: https://stackoverflow.com/questions/7144843/extra-space-under-textarea-differs-along-browsers

ryandialpad (approved):
w/ minor comments

Great Work 🚀 
 
The attached gif is pretty neat, would make for a very interesting input box :)

dialpad-drew:
> Found a better solution here that also includes a good explanation: https://stackoverflow.com/questions/7144843/extra-space-under-textarea-differs-along-browsers

Good catch, man. Thanks for handling this! 🙌

PR #295, opened 2021-04-27
Design documentation icons

dialpad-drew:
## Description
* Updates all icon documentation:
  * Brand Icon page
  * System Icon page
  * Weather Icon page
  * Icon size page
* Adds an icon page grid class and uses existing icon classes; I initially tried moving all of this to inline utility classes and it was going to be too much re-work for something that already existed
* Creates an icon-page layout that works for Brand and System icons; unfortunately it doesn't quite work for Weather icons because of the page subsections
* Re-organizes some of the Design files so they're nested in subdirectories like Utilities
* Fixes a defect in which Design's side nav didn't properly show active pages

**To test**
* Look through Design: Icons pages to ensure they look as expected and side navigation shows the proper active page


ryandialpad (approved):
LGTM 🚀

PR #299, opened 2021-04-28
V6 component button docs

dialpad-drew:
## Description
Updates Button docs. There's some additional work that will need to be done here, but that's for another ticket.


ryandialpad (approved):
Great work 🚀 

Should the muted button variant have any hover styles?

Visually the only indication I get is the cursor style changing and that might be difficult for individuals with vision deficiencies (I know it was hard enough for myself to notice)

dialpad-drew:
> Great work 🚀
> 
> Should the muted button variant have any hover styles?
> 
> Visually the only indication I get is the cursor style changing and that might be difficult for individuals with vision deficiencies (I know it was hard enough for myself to notice)

I think it does, but it's the same color as the background-color. I _was_ going to make the hover a color with alpha instead, but didn't because I learned rectangular Muted buttons aren't a real thing. They were in the Figma file I worked from and so I did them, but they aren't in DT5 CSS, Vue, or in our product. So, I have a ticket coming up soon for cleaning up things like that.

In short: I'm gonna pull them out.

PR #301, opened 2021-04-29
Undoing build/ and dist/ pathing changes from early December.

dialpad-drew:
## Description
This is undoing the `build/` and `dist/` pathing changes that were committed on Dec 2, 2020. Those changes would require updates in Handset and across UC and UV, and is just one more thing we need to update and manage when implementing`6.0`. Undoing them is an attempt to simplify, even if only slightly, the current v6 CSS=>Vue work.

These paths should now match DT5.x.

I would like to re-visit this and have a more standard directory structure as was intended, but at a later date.


ryandialpad (approved):
Thanks Drew!

PR #308, opened 2021-05-01
Clean up Getting Started documentation

dialpad-drew:
## Description
- Updates "Getting Started" documentation
- Re-writes "Using Dialtone"

## Obligatory GIF
tenor


ryandialpad (approved):
Looks good to me!

BTW, users outside of Dialpad will not be able to access handset currently until / if it is open sourced

PR #309, opened 2021-05-04
Re-factored muted buttons

dialpad-drew:
## Description
I removed the rectangular Muted Button style as I now realize that's not a DT5 style despite being in the Figma file. It is, however, the base style for circle buttons, so I nested the muted styles inside `d-btn--circle` and had to add `!important` to danger and inverted button styles. I really hate having to do that, but it was either that or re-declaring some of the hover and focus styles inside `d-btn--circle`.

In the long run, I'm not sure I see a reason in which circle buttons have different base styles than non-circle buttons. We should look at this in the future and consider cleaning it up if possible.

I also re-factored the icon button styles so their semantics would be more inline with how we include icons in other components (see input). This would be a breaking change with the Vue component, which I believe uses `.d-btn__icon--last`.


docs/assets/css/dialtone.css
```diff
@@ -760,80 +765,78 @@ body {
...
-}
-.d-btn--muted:focus {
-  --button--bs-color-outer: var(--black-300);
-}
-.d-btn--muted.d-btn--outlined {
-  border-color: var(--black-500);
-}
 .d-btn--danger {
-  --button--fc: var(--red-400);
+  --button--fc: var(--red-400) !important;
```
ryandialpad, line 769: Quick question @dialpad-drew 

Could you explain why we are adding `!important` to the dialtone styles?

lib/build/less/components/button.less
```diff
@@ -180,67 +184,46 @@
 //  $$  MUTED BUTTON
 //  ----------------------------------------------------------------------------
 .d-btn--muted {
```
ryandialpad: Did we want to remove the class entirely?

lib/build/less/components/button.less
```diff
@@ -180,67 +184,46 @@
...
-        // }
-    }
 }
 
 //  $$  DANGER BUTTON
 //  ----------------------------------------------------------------------------
 .d-btn--danger {
 
-    --button--fc: var(--red-400);
+    --button--fc: var(--red-400) !important;
```
ryandialpad: Quick question for @dialpad-drew 

Could you explain why we are adding `!important` to the dialtone styles?

lib/build/less/components/button.less
dialpad-drew: yes

lib/build/less/components/button.less
```diff
@@ -292,22 +275,44 @@
...
+    }
+
+    &:focus {
+        --button--bs-color-outer: var(--black-300);
+    }
+    // &:active {
+    //     border-color: hsla(var(--black-700-h) var(--black-700-s) var(--black-700-l) ~" / " 50%);
+    //     background-color: hsla(var(--black-700-h) var(--black-700-s) var(--black-700-l) ~" / " 10%);
+    // }
+
```
ryandialpad, line 302: @dialpad-drew Are we going to come back and update all of these soft deleted styles?

ryandialpad (approved):
w/ 2 questions

lib/build/less/components/button.less
dialpad-drew: Yes. I didn't realize it that muted buttons weren't a real thing; they were in the original Figma file, so I designed them for the DT6 file. It wasn't until reviewing Brad's button CSS PR that I realized there's no such class or rectangular style as `d-btn--muted` in DT5. And I certainly don't want to unintentionally introduce a style.

lib/build/less/components/button.less
dialpad-drew: I explained in the PR message, but will repeat here in case it didn't make sense.

We either need to add `!important`, which, again, I'd prefer _not_ to do, or repeat those styles inside `d-btn--circle`. Otherwise, they get partially overwritten by `d-btn--circle`'s styles because it has more specificity.

If you have another way to achieve the result, I'm sure I'd prefer it to adding `!important`. Take a look at it and let me know.

lib/build/less/components/button.less
dialpad-drew: Yup. Already answered this about these specific styles in a different PR.

lib/build/less/components/button.less
dialpad-drew: Now that we're talking about it, maybe I _should_ duplicate the styles. I _really_ hate using `!important` and I wonder if adding it here means it could collide with any additional utility classes added to button.

dialpad-drew: Looking at DT5, it appears that the original circle buttons repeated their `:hover` and `:focus` styles.

docs/_data/button.json
```diff
@@ -73,12 +68,17 @@
...
+      "description": "Base style for including an icon with a label."
+    },
+    {
+      "class": "d-btn__icon--left",
+      "applies": "d-btn__icon",
       "description": "Positions the icon to the left of the text label."
     },
     {
-      "class": "d-btn__icon--last",
+      "class": "d-btn__icon--right",
```
braddialpad, line 79: Do we keep track of these types of breaking class name changes?

lib/build/less/components/button.less
braddialpad: while duplicate styles suck, I think I'd prefer them over !important

lib/build/less/components/button.less
dialpad-drew: Ahh, I didn't understand your question nor did I realize I left the class in there. Yes, we do. Removing now. 🤕

docs/_data/button.json
dialpad-drew: My understanding is @ryandialpad was noting them in the PRs and was going to go back and aggregate them all. If you want to trap it now so we don't forget it, maybe we put it in the Button contractor story Ryan specked out when we walked the components a few weeks back....?

lib/build/less/components/button.less
dialpad-drew: Fair enough. Made the change.

lib/build/less/components/button.less
dialpad-drew: Would you prefer if I removed them for now and just add them back when we add back `:active` styles?

ryandialpad: I am fine leaving them soft deleted if we plan to update them in the short term. If it's something longer term than removing them would be better.

PR #319, opened 2021-05-14
More shortcode clean-up

### LINKED ISSUE #321: Documentation typoes
## Button definition:
A button is a UI element which **_allow_** users to take an action throughout the app. It is important a button is identifiable, consistent, and communicates its actions clearly, and is appropriately sized to its action.

## Input definition:
An input field is an input control that allows users to enter **_alphanumberic_** information. It can have a range of options and supports single line and multi-line lengths, as well as varying formats, including numbers, masked passwords, etc.

dialpad-drew:
## Description
More miscellaneous clean-up. Though it touches a bunch of files, this PR is really simple and only really contains a few small changes:
* Replaces `ul` and `ol` tags (except for those used in examples) with shortcodes so their class declarations are centralized and easier to manage
* Replaces `p` tags (except for those used in examples) with the paragraph shortcode for the same reason as the above bullet
* Removes max-width (`.d-w75ch`) class from paragraph shortcode; paragraphs shouldn't need to manage its own width
* Replaced some Greek copy
* Standardized some spacing
* Fixed a couple areas of low contrast
* Fixes #321


ryandialpad (approved):
Great work adding these new shortcodes!

PR #324, opened 2021-05-18
Link polish

dialpad-drew:
## Description
This PR polishes link and button style documentation a bit:
1. Hides disabled link styles from link documentation. There's not really such a thing as a disabled link in semantics or our codebase.
2. Adds disabled link styles in button documentation. This is because a disabled link style was intended to be used with a `<button>`.
3. Adds the `disabled` attribute to disabled link styles so it will target a properly disabled `<button>`.
4. Removes the `.d-link--inverted-danger` style. This was creating awkward issue trying to standardize and organize storybook props for components with inverted styles. We'll re-add it when we add dark mode and/or inverted color support more holistically. **Note:** this class isn't currently in use in our products.
5. Removes unused, duplicated button documentation.


ryandialpad (approved):
:shipit: 

Great Work!!

I am assuming we are keeping `d-link--disabled` for legacy support in addition to the suggested usage of `disabled`

braddialpad:
> :shipit:
> 
> Great Work!!
> 
> I am assuming we are keeping `d-link--disabled` for legacy support in addition to the suggested usage of `disabled`

We're actually probably going to end up using it INSTEAD of disabled due to this article David posted yesterday.

https://css-tricks.com/making-disabled-buttons-more-inclusive/

ryandialpad:
> > :shipit:
> > Great Work!!
> > I am assuming we are keeping `d-link--disabled` for legacy support in addition to the suggested usage of `disabled`
> 
> We're actually probably going to end up using it INSTEAD of disabled due to this article David posted yesterday.
> 
> https://css-tricks.com/making-disabled-buttons-more-inclusive/

I still need to read this article, thanks for the heads up 🙌

dialpad-drew:
> > :shipit:
> > Great Work!!
> > I am assuming we are keeping `d-link--disabled` for legacy support in addition to the suggested usage of `disabled`
> 
> We're actually probably going to end up using it INSTEAD of disabled due to this article David posted yesterday.
> 
> https://css-tricks.com/making-disabled-buttons-more-inclusive/

Are we doing that for all buttons right now? The CSS doesn't support that as it doesn't offer styles for `[aria-disabled=true]` nor does the CSS documentation reflect that approach.

ryandialpad:
> > > :shipit:
> > > Great Work!!
> > > I am assuming we are keeping `d-link--disabled` for legacy support in addition to the suggested usage of `disabled`
> > 
> > 
> > We're actually probably going to end up using it INSTEAD of disabled due to this article David posted yesterday.
> > https://css-tricks.com/making-disabled-buttons-more-inclusive/
> 
> Are we doing that for all buttons right now? If so, I need to update the a11y documentation.

cc @braddialpad 

Halfway through reading the article and it is clear to me using `disabled` is a bad idea for accessibility.

braddialpad:
> Are we doing that for all buttons right now? The CSS doesn't support that as it doesn't offer styles for `[aria-disabled=true]` nor does the CSS documentation reflect that approach.

Yeah, we can make this change in a future update. Not super necessary right now

PR #326, opened 2021-05-24
Added Amex, Mastercard, and Visa brand logos.

dialpad-drew:
## Description
Adds Amex, Mastercard, and Visa brand icons per Vicky's request.

## Pull Request Checklist

 - [X] Ask the contributors if a similar effort is already in process or has been solved.
 - [X] Review the [contribution guidelines](https://github.com/dialpad/dialtone/blob/staging/.github/CONTRIBUTING.md).
 - [X] Use `staging` as your pull request's base branch. (All PRs using `production` as its base branch will be declined).
 - [X] Ensure all `gulp` scripts successfully compile.
 - [X] Update, remove, or extend all affected documentation.
 - [X] Provide a description what is being changed, extended, or removed.
 - [X] Link to any issue(s) this request is resolving.
 - [X] Ensure no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).
 - [X] Request a review from Drew Chandler, Joshua Hynes, or Ted Goas.

PR #340, opened 2021-05-28
Added Amex, Mastercard, and Visa brand icons from DT5

dialpad-drew:
## Description
Adds the Amex, Mastercard, and Visa brand icons to DT6 as they were just added to DT5 in PR #326


ryandialpad (approved):
:shipit:

PR #346, opened 2021-06-07
Removing filled radio style on focus

### LINKED ISSUE #343: Radio button focus styles should not impact the checked state
It is [possible](https://www.w3.org/wiki/RadioButton), per W3C, to focus on a radio button without it being checked (for example if you arrive at a group that has no checked button, or if we programmatically focus it with JS). Right now our `:focus` styles make it visually look checked even though it's not:

Jun-02-2021 17-25-55

We need to change these styles so a focused radio button appears with a focus ring outline but does not affect the visually checked state. Example from Carbon:

Jun-02-2021 17-22-47

ryandialpad:
## Description

Updating radio styles so that the focus styles do not look selected

Closes #343 

## Pull Request Checklist

## Screenshot

Screen Shot 2021-06-07 at 9 55 11 AM


ryandialpad:
@dialpad-drew We will have to back port these changes to DT5 as well

lib/build/less/components/radio-checkbox.less
```diff
@@ -262,7 +262,6 @@
 
         &:focus,
         &:checked:focus {
-            --check-radio--bgc:     var(--check-radio--color);
```
dbecherdialpad, line 265: is this the only real change?

lib/dist/css/dialtone.css
```diff
@@ -35518,7 +35519,8 @@ ul {
   text-decoration: underline !important;
 }
 .d-td-dotted {
-  text-decoration: underline dotted !important;
+  -webkit-text-decoration: underline dotted !important;
```
dbecherdialpad: do you know why all this was added? I don't think these vendor prefixes are needed.

lib/build/less/components/radio-checkbox.less
ryandialpad: Yup

lib/dist/css/dialtone.css
ryandialpad: Idk tbh, I will have to look into it further

dialpad-drew: @ryandialpad when's the last time you updated ~package~ project dependencies? I seem to recall a recent dependency update removing some of these prefixes.

dialpad-drew: Ahh yes. Here: https://github.com/dialpad/dialtone/pull/333/files

dialpad-drew (approved):
Would recommend updating project dependencies and re-compiling CSS. Should remove some of these diffs.

Otherwise, LGTM. :shipit:

package-lock.json
```diff
@@ -12393,7 +12393,8 @@
     },
     "y18n": {
       "version": "3.2.1",
-      "resolved": "",
+      "resolved": "https://registry.npmjs.org/y18n/-/y18n-3.2.1.tgz",
+      "integrity": "sha1-bRX7qITAhnnA136I53WegR4H+kE=",
```
ryandialpad, line 12397: Looks like last time we ran an install this dep failed to resolve

Child of `yargs` which is a child of `browser-sync` which is a child of `eleventy`

dbecherdialpad (approved):

PR #367, opened 2021-07-14
Updating button documentation to include d-btn--disabled

dialpad-drew:
## Description
Updates button documentation to mention `d-btn--disabled`.

### DIFF
--- a/docs/components/button.html
+++ b/docs/components/button.html
@@ -124,14 +124,23 @@ description: A button is a UI element which allows users to take an action throu
     </div>
     <div class="d-stack8">
       {% header "h3", "Disabled" %}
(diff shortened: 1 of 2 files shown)


docs/components/button.html
```diff
@@ -124,14 +124,23 @@
     </div>
     <div class="d-stack8">
       {% header "h3", "Disabled" %}
+      {% paragraph %}Buttons can be disabled using either the {% code %}disabled{% endcode %} attribute or a Dialtone class. Use the attribute when a button should appear disabled and not recieve focus; use the class when a button should appear disabled but still recieve focus (i.e. a disabled button with a tooltip). Because using the disabled class requires additional implementation to prevent the click handler, it's highly recommended you use the <a class="d-link" href="https://vue.dialpad.design/?path=/story/elements-button--default" target="_blank">button Vue component</a>.{% endparagraph %}
```
dbecherdialpad: nit: the vue component doesn't (yet) handle this either. The `disabled` prop just sets the `disabled` html attribute

dbecherdialpad (approved):
Maybe mention this in the Accessibility section as well? Basically the tldr of that article (use `aris-disabled` and manually prevent the click handler if using the disabled class)

docs/components/button.html
dialpad-drew: Fair point. I just pushed an update that removes the mention of using Vue to solve for this and adds a bullet to the a11y section

PR #379, opened 2021-07-27
Fixes issue in which pink-700 didn't generate correct CSS classes

### LINKED ISSUE #345: Secret pink color stop
The dialtone v6 docs seem to contain a hidden color stop for Pink (pink-700):

Screen Shot 2021-06-02 at 1 00 10 PM

I bet there's a variable that's undefined and causing the background color to be invalid there.

dialpad-drew:
## Description
This PR fixes #345 , which highlights an issue in which the `pink-700` color stop didn't have the properly generated styles. Pink -700's border, background, font, etc. color styles were undefined because `pink-700` hadn't been properly identified in the Less file that generates these. The mixin only loops through colors stops 100-600, and any above or below that range need to be manually identified.


ryandialpad (approved):
Good catch!

PR #382, opened 2021-08-02
V6 favicon

dialpad-drew:
## Description
Updated the DT6 doc site favicons (from blue to purple) and the implementation method. Also supports device dark and light mode. @hynes-dialpad are the screenshots below in line with your expectations when you were looking at this?

### Chrome
Screen Shot 2021-08-02 at 10 43 54 AM
Screen Shot 2021-08-02 at 10 44 01 AM

### Safari
Screen Shot 2021-08-02 at 10 44 29 AM
Screen Shot 2021-08-02 at 10 44 36 AM

### Firefox
Screen Shot 2021-08-02 at 10 46 26 AM
Screen Shot 2021-08-02 at 10 46 31 AM

### Edge
Screen Shot 2021-08-02 at 10 52 37 AM
Screen Shot 2021-08-02 at 10 52 42 AM


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@dialpad/dialtone",
   "description": "The design system for Dialpad and UberConference",
-  "version": "6.0.0-beta.10",
+  "version": "6.0.0-beta.11",
```
ryandialpad, line 4: Do we bump the version just before release? Because we are currently on beta 11 for Dialtone V6

docs/assets/images/favicons/site.webmanifest
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "",
+  "short_name": "",
```
ryandialpad:
```suggestion
  "name": "Dialtone",
  "short_name": "Dialtone",
```

We should at the very least have these filled out for our site manifest

ryandialpad: Maybe also look into adding back `description` and `lang` at some time

ryandialpad (approved):
w/ comment and question

package.json
dialpad-drew: I think it's because I don't push the version bump up to GH. I push the tags and I publish the npm package, but it's not pushed to origin. I should probably do that.

hynes-dialpad:
@dialpad-drew The Edge dark mode doesn't look correct. Should it be white instead of purple? Otherwise LGTM!

PR #384, opened 2021-08-02
Fix hero image size bug in Safari

dialpad-drew:
## Description
Fixes the hero image sizing issue in Safari. This usually occurs when an image's parent uses `flex` and the image has a declared `width`. The implied `height: auto` doesn't always work as expected.

### Before
Screen Shot 2021-08-02 at 10 55 48 AM

### After
Screen Shot 2021-08-02 at 10 55 45 AM


tedgoas-dialpad (approved):
✅ The image fix look good.

We probably want to improve upon the _whole_ page's responsive view, right?
Screen Shot 2021-08-02 at 3 45 19 PM

dialpad-drew:
> ✅ The image fix look good.
> 
> We probably want to improve upon the _whole_ page's responsive view, right?
> Screen Shot 2021-08-02 at 3 45 19 PM

Correct. In fact, I just pushed some fixes to the hero (nav is a whole other story). It's still not super great, but it's better. However, to achieve this you'll see this required me to move _away_ from utility classes into custom CSS because DT6 offers limited media query support and the class's immutability gives them priority in the cascade, which meant I couldn't use utility and then override with a media query. 😬

docs/assets/less/dialtone-docs.less
```diff
@@ -273,15 +274,18 @@
 }
 
 .hero--inner-wrapper {
-  padding-top: 12.8rem;
   padding-bottom: 12.8rem;
+  padding-top: 12.8rem;
   @media (max-width: 640px) {
-    flex-direction: column;
+    flex-direction: column
```
ryandialpad:
```suggestion
    flex-direction: column;
```

Looks like we accidentally deleted a semi colon here

ryandialpad (approved):
w/ comment

docs/assets/less/dialtone-docs.less
dialpad-drew: 👀 👍

PR #392, opened 2021-08-06
V6 add version

dialpad-drew:
## Description
Adds CSS and Vue version badges to the home page.

### Background
Dialtone 5's version was stored in a HTML partial (`version.html`) and included in the site nav. There was a gulp task `gulp version` that would bump that partial to match whatever the version had been updated to in `package.json`.

This PR removes that while adding two version badges to the Dialtone home page. The CSS badge simply reflects the version number from the GitHub project. No more manual bumping on the doc site! 🙌 

The Vue version is hard-coded because the Dialtone-Vue project is not public. Therefore, we'll have to manually manage this for the time-being unless there's a smarter solution. I'm all 👂 if so.


gulpfile.js
```diff
@@ -639,10 +639,10 @@ exports.fonts = series(
...
 
 //  --  UPDATES DIALTONE VERSION
-exports.version = series(
-    docVersion,
-    commitDocVersion
-);
+// exports.version = series(
+//     docVersion,
+//     commitDocVersion
+// );
```
ryandialpad: If this is no longer needed we can probably just remove the code instead of the soft deletion.

ryandialpad (approved):
w/ comment

PR #393, opened 2021-08-09
Dt 44 refactor button component classes

juliodialpad:
## Description
We need to refactor our button component classes because they are used as Less mixins and will not be resolved unless the class name is completely defined in the Dialtone Less files

https://switchcomm.atlassian.net/browse/DT-44?atlOrigin=eyJpIjoiNGYyNzViYTc4Yjk4NDkyY2I0MDk0NTk5ZmVlZjBlZTMiLCJwIjoiaiJ9

## Pull Request Checklist

### DIFF
--- a/lib/build/less/components/button.less
+++ b/lib/build/less/components/button.less
@@ -148,17 +148,17 @@
     margin-bottom: var(--sun4);
     width: 1em;
     height: 1em;
+}
 
-    &.d-btn__icon--left {
-        margin-left: var(--sun2);
-        margin-right: var(--su4);
-    }
+.d-btn__icon--left {
+    margin-left: var(--sun2);
+    margin-right: var(--su4);
+}
 
-    &.d-btn__icon--right {
-        order: 1;
-        margin-left: var(--su4);
-        margin-right: var(--sun2);
-    }
+.d-btn__icon--right {
+    order: 1;
+    margin-left: var(--su4);
+    margin-right: var(--sun2);
 }
 
 
--- a/lib/build/less/components/button.less
+++ b/lib/build/less/components/button.less
@@ -8,9 +8,10 @@
 //
 //  TABLE OF CONTENTS
 //  • BASE STYLE
+//  • STATES
+//  • SIZES
 //  • VISUAL STYLES
 //  • ALTERNATE STYLES
-//  • SIZES
 //
 //  ============================================================================
 //  $   BASE STYLE
@@ -72,12 +73,29 @@
   .d-svg {
     .d-svg--size16();
   }
+}
 
-  //  --  LOADING
-  //  ------------------------------------------------------------------------
-  &.d-btn--loading {
-    .d-btn-loading();
-  }
+//  ============================================================================
+//  $   STATES
+//  ============================================================================
+//  $$  LOADING
+//  ------------------------------------------------------------------------
+.d-btn--loading {
+  .d-btn-loading();
+}
+
+//  $$  DISABLED
+//  ----------------------------------------------------------------------------
+.d-btn--disabled {
+  pointer-events: none;
+}
+[disabled],
+.d-btn--disabled {
+  color: var(--black-400);
+  background-color: var(--black-100);
+  cursor: not-allowed;
+  border-color: transparent;
+  transition: none;
 }
 
 //  ============================================================================
@@ -95,6 +113,9 @@
     height: 1.2rem !important;
   }
 }
+.d-btn--circle.d-btn--xs {
+  padding: calc(var(--su8) - var(--su1));
+}
 
 //  $$  SMALL
 //  ----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


lib/build/less/components/button.less
```diff
@@ -8,147 +8,158 @@
...
+  pointer-events: none;
+}
+[disabled],
+.d-btn--disabled {
+  color: var(--black-400);
+  background-color: var(--black-100);
+  cursor: not-allowed;
+  border-color: transparent;
+  transition: none;
+}
```
ryandialpad: This will apply styles to any element with `disabled`

Example:
```html
<input id="my-radio" type="radio" disabled />
<label for="my-radio">My radio</label>
```

The radio would have all of the styles listed above.

Your best bet for refactoring this would be to create an internal mixin which is used for the nested `[disabled]` selector and the disabled btn class.

lib/build/less/components/button.less
```diff
@@ -8,147 +8,158 @@
...
+
+  .d-svg {
+    // no icon-size 12 exists so setting manually
+    width: 1.2rem !important;
+    height: 1.2rem !important;
+  }
+}
+.d-btn--circle.d-btn--xs {
+  padding: calc(var(--su8) - var(--su1));
 }
```
ryandialpad: We should revert the changes to the circle class refactor as we only need the root classes to be un-nested so that they can be resolved. The nested selectors will work as intended.

Example:
```html
<button class="d-btn--circle d-btn--xs" />
```

```less
.my-custom-btn-class {
  .d-btn--circle;
  .d-btn--xs;
}
```

ryandialpad (changes requested):
* Revert changes to circle btn size un-nesting
* Move `[disabled]` styles back in `d-btn` to prevent any element with `disabled` from being selected with those styles

juliodialpad (commented):
Reverted the un-nesting and moved [disabled] states

lib/build/less/components/button.less
```diff
@@ -8,147 +8,149 @@
...
-        font-size: var(--fs24);
-        line-height: var(--lh8);
-
-        .d-svg {
-            .d-svg--size24();
-        }
-    }
+//  $$  DISABLED
+//  ----------------------------------------------------------------------------
+.d-btn {
```
ryandialpad: `d-btn` is defined twice, could we please move these into the existing `d-btn` class above?

lib/build/less/components/button.less
```diff
@@ -190,191 +192,205 @@
...
+}
+.d-btn__icon--left {
+  margin-left: var(--sun2);
+  margin-right: var(--su4);
+}
+.d-btn__icon--right {
+  order: 1;
+  margin-left: var(--su4);
+  margin-right: var(--sun2);
 }
```
ryandialpad:
```suggestion
.d-btn__icon {
  display: flex;
  margin-top: var(--sun4);
  margin-bottom: var(--sun4);
  width: 1em;
  height: 1em;
}

.d-btn__icon--left {
  margin-left: var(--sun2);
  margin-right: var(--su4);
}

.d-btn__icon--right {
  order: 1;
  margin-left: var(--su4);
  margin-right: var(--sun2);
}
```

Ensure that you have padding between class definitions

ryandialpad: Same goes with circle button size selectors

lib/build/less/components/button.less
ryandialpad: `d-btn--disabled` should also be defined at root level so it can be resolved

Example
```less
.d-btn {
  ...
  &[disabled] {
    color: var(--black-400);
    background-color: var(--black-100);
    cursor: not-allowed;
    border-color: transparent;
    transition: none;
  }
  ...
}

.d-btn--disabled {
  pointer-events: none;
  color: var(--black-400);
  background-color: var(--black-100);
  cursor: not-allowed;
  border-color: transparent;
  transition: none;
}
```

You can build on this example if you want to make the following styles a little more re-usable

```less
color: var(--black-400);
background-color: var(--black-100);
cursor: not-allowed;
border-color: transparent;
transition: none;
```

lib/build/less/components/button.less
ryandialpad: and the button icon classes

ryandialpad (changes requested):
See comments

lib/build/less/components/button.less
```diff
@@ -131,23 +143,23 @@
...
+    &.d-btn__icon--left {
+        margin-left: var(--sun2);
+        margin-right: var(--su4);
+    }
 
+    &.d-btn__icon--right {
+        order: 1;
+        margin-left: var(--su4);
+        margin-right: var(--sun2);
+    }
```
ryandialpad: These still need to be un-nested to be resolved as mixins

ryandialpad (approved):
w/ one remaining change

See comments

PR #407, opened 2021-08-20
refactor: dt-31 Modal transition refactor to use Vue transition compatible CSS classes

### LINKED ISSUE #159: Modal transition: refactor to use Vue-transition-compatible CSS classes instead of aria-hidden
**Summary**
The current modal styling is controlled by the setting of the `aria-hidden` attr.  This means that the modal is always rendered in the DOM, even if it's never triggered and made visible.

If the transition was instead accomplished with a set of css classes for the various transition-stages (e.g. animate-in, animate-out), we could map those directly to Vue's transition classes: https://v3.vuejs.org/guide/transitions-enterleave.html#transition-classes.

**Details**
Per discussion w/ @dbecherdialpad on the Handset Modal PR https://github.com/dialpad/firespotter/pull/7210

> I am more convinced that this HsModal component should use LazyShow as its root node so that we don't render the modal content until it's open. It's fine for now but you can imagine a scenario where we have a lot of modals on a page (maybe even one per row in a table!) and it will be pretty inefficient to render them all into the DOM at once instead of when they open.

> The only real change that would be required to do this would be to use a CSS class for the transition instead of the aria-hidden prop.

And from earlier discussion w/ Ted/David in slack:

Something like:
```
.d-modal--animate { sets the transition properties... }
.d-modal--animate-in { transitions to visible }
.d-modal--animate-out { transitions to hidden }
```

Where the various vue transition classes could then be bound to those in the modal component, e.g. `fade-enter-active: d-modal--animate-in`.

josedialpad:
## Description
Issue: https://switchcomm.atlassian.net/browse/DT-31
This PR adds modal transition classes to apply during transition instead of relying on `aria-hidden` prop to transition.

Updated documentation modal examples in order to use these classes.

## Pull Request Checklist

## Obligatory GIF
CleanShot 2021-08-20 at 10 59 51

### DIFF
--- a/lib/build/less/components/modal.less
+++ b/lib/build/less/components/modal.less
@@ -44,10 +44,11 @@
 
     background-color: var(--modal--bgc);
 
-  visibility: visible;
-  opacity: 1;
-  z-index: var(--zi-modal);
-  transform: translate3d(0,0,0) scale3d(1,1,1);
+    visibility: visible;
+    opacity: 1;
+    z-index: var(--zi-modal);
+    transform: translate3d(0,0,0) scale3d(1,1,1);
+    will-change: visibility, z-index, opacity;
 
     @media (prefers-reduced-motion) {
         transition: none;
@@ -77,6 +78,7 @@
   opacity: 1;
   z-index: var(--zi-modal);
   transform: translate3d(0,0,0) scale3d(1,1,1);
+  will-change: visibility, z-index, opacity, transform;
 
   .d-stack16();
 }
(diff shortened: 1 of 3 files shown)


dialpad-drew:
How does this impact the DT6 implementation? Anything we need to note there?

Would we need to add these new classes to the class table?

josedialpad:
@dialpad-drew since with this PR the transition is made with CSS classes, in the DT6 implementation we need to apply the classes during the transition. In the [following PR](https://github.com/dialpad/dialtone-vue/pull/137) the Vue transition wrapper was used with these new transition classes.

> Would we need to add these new classes to the class table?

Yes, and we could add some documentation about it in the dialtone documentation.
Update: added transition classes in Modal classes documentation table https://github.com/dialpad/dialtone/commit/e052c1ca3a72496a18ba5cae47ff7adcad6fc4d3

dialpad-drew (approved):

dialpad-drew:
Are we waiting for David's approval?

josedialpad:
@dialpad-drew ideally yes. I'd like his review especially in the implementation PR https://github.com/dialpad/dialtone-vue/pull/137 but If we're good with these changes we can merge this and just wait for the review on the implementation PR (probably after Dialtone 6 migration since it's the priority).

lib/build/less/components/modal.less
```diff
@@ -49,11 +49,6 @@
...
     backface-visibility: hidden;
 
-    transition:
-        opacity 100ms var(--ttf-in-out) 10ms,
-        z-index 0s 200ms,
-        visibility 0s 200ms,
-        transform 100ms var(--ttf-in-out) 10ms;
     will-change: visibility, z-index, opacity;
 
     @media (prefers-reduced-motion) {
```
dbecherdialpad: I think this `prefers-reduced-motion` query needs to be wrapped around the new transition classes below (and make sure the classes work with `prefers-reduced-motion` enabled)

dbecherdialpad (changes requested):

lib/build/less/components/modal.less
josedialpad: Thanks, you're right. Updated to put `prefers-reduced-motion` in the transition classes. https://github.com/dialpad/dialtone/pull/407/commits/17ff5c43ac3dcc33de9998f30820e0bd4d7525b1

> and make sure the classes work with prefers-reduced-motion enabled

The classes work but not the transition as expected.

dbecherdialpad (approved):

PR #409, opened 2021-08-25
Removed universal declaration of border-box

juliodialpad:
## Description
Removed universal declaration of border-box in dialtone-globals and added to dialtone-docs

https://switchcomm.atlassian.net/browse/DT-92?atlOrigin=eyJpIjoiZmJiODE1MDExZGRjNDQxNDhkNzUyMDE0NWE4MWMwYjYiLCJwIjoiaiJ9

## Pull Request Checklist


hynes-dialpad:
A few quick thoughts:

1. If we're removing `border-box` from Dialtone globals, but adding it into the doc less, will this affect our ability to check if components and styles are working correctly if the global styles is set to `content-box`?
2. Do we need to re-add `border-box` to each component? The reason I added this to the globals file was so we could reduce redeclaring this within each component.

On my previous design system we had a similar situation and our solution was to make a mixin that a project could determine using a Less mixin. I could dig up an example if you're interested.

dbecherdialpad (approved):

hynes-dialpad:
@dbecherdialpad I see you approved this PR. Any thoughts on the concerns I shared above?

dbecherdialpad:
@hynes-dialpad yeah we can't set border-box globally, in UV at least, because it messes up a bunch of the old CSS. Currently Dialtone 5 does not provide set this globally either, and both UC and UV at least set it themselves in the appropriate place (UC globally, UV for every vue component). I think it's fair for now to just say that any project using dialtone has to set this.

On possibility would be to just make this a compile-time option similar to what we do for other global things like fonts, which would solve the UV use case as well.

hynes-dialpad:
@dbecherdialpad That's fine. My point though is that we should make sure components still have a declared `box-sizing: border-box` in them if they do not already. I _may_ have removed them in my initial D6 work earlier this year because I assumed it would be globally set.

dbecherdialpad:
@hynes-dialpad I don't think we need to (however I checked and I still see it at least in some dialtone v6 components, maybe @dialpad-drew can confirm). I think it's enough to just say "you must set this yourself on some parent of the dialtone component" since like 99% of sites are doing that anyway. If anything looks funky with the dialtone components in UV I'll know it's bc of border-box :)

dialpad-drew:
Yea, it looks like it's added on the parent class for most, though not all, components. It see it on the following:

* `d-banner`
* `.d-breadcrumbs > ol`
* `.d-btn`
* `.d-label`
* `.d-description`
* `.d-input`
* `.d-textarea`
* `.d-link`
* `.d-notice`
* `.d-select`
* `.d-table`
* `.d-tab`
* `.d-toast`

It looks like the following don't have this declared:
* Avatar
* Badge
* Checkbox (declared in its label and description text)
* Modal
* Radio (declared in its label and description text)
* Tooltip

Am I understanding correctly we'd want to add `border-box` to those parent classes?

hynes-dialpad:
@dialpad-drew Up to you. If we don't have `border-box` declared for the D5 components you listed above I guess this won't make a big difference. Sorry for creating confusion here.

hynes-dialpad:
And thanks for checking @dialpad-drew! 👊

dialpad-drew:
> @dialpad-drew Up to you. If we don't have `border-box` declared for the D5 components you listed above I guess this won't make a big difference. Sorry for creating confusion here.

No worries, dude! Better to know than not. 

A quick look at DT5 suggests `border-box` isn't on those six components (avatar, badge, etc.). I think we should be good. 👍  😄

PR #412, opened 2021-08-27
Added UC toast shims

juliodialpad:
## Description
Dialpad Meetings uses three custom toast styles that Dialtone needs to continue supporting. We need to add shims for the following classes:

.d-toast--chat
.d-toast--organizer
.d-toast--viewing

https://switchcomm.atlassian.net/browse/DT-95?atlOrigin=eyJpIjoiM2I1N2NhMjYzNTI5NGYzODg4ZTE2ZjgyZDhlNDNkZGIiLCJwIjoiaiJ9

## Pull Request Checklist


lib/build/less/components/toast.less
```diff
@@ -134,3 +134,65 @@
...
+//  $$  VIEWING
+//  ----------------------------------------------------------------------------
+.d-toast--viewing {
+ //
+}
+
+.d-toast--organizer,
+.d-toast--viewing {
+  &:before {
+    .d-bgc-pink-500();
```
dialpad-drew: I'm not sure if we're trying to follow things pattern anymore. I _thought_ we were trying to use CSS vars as much as possible.

lib/build/less/components/toast.less
```diff
@@ -134,3 +134,65 @@
...
+  .d-notice__icon {
+    color: @pink-500;
+  }
+
+  &:before {
+    content: "";
+    position: absolute;
+    top: -(@su1);
+    right: -(@su1);
+    left: -(@su1);
```
dialpad-drew: Use CSS vars instead of Less vars.

This should work: `var(--sun1)`

lib/build/less/components/toast.less
```diff
@@ -134,3 +134,65 @@
...
+}
+
+.d-toast--organizer,
+.d-toast--viewing {
+  &:before {
+    .d-bgc-pink-500();
+  }
+
+  .d-notice__icon {
+    color: @pink-500;
```
dialpad-drew: CSS var

lib/build/less/components/toast.less
```diff
@@ -134,3 +134,65 @@
...
+
+  &:before {
+    content: "";
+    position: absolute;
+    top: -(@su1);
+    right: -(@su1);
+    left: -(@su1);
+
+    height: @su6;
+    border-radius: @br4 @br4 0 0;
```
dialpad-drew: CSS vars

lib/build/less/components/toast.less
```diff
@@ -134,3 +134,65 @@
         --toast--fc:       var(--fc-dark);
     }
 }
+
+//  $$  CHAT
+//  ----------------------------------------------------------------------------
+.d-toast--chat {
+  .d-notice__icon, .d-toast__meta {
+    color: @black-600;
```
dialpad-drew: CSS var

lib/build/less/components/toast.less
dialpad-drew: What's the difference doing this and putting a `border-top` on it?

lib/build/less/components/toast.less
```diff
@@ -134,3 +134,65 @@
...
+  }
+
+  &:before {
+    content: "";
+    position: absolute;
+    top: -(@su1);
+    right: -(@su1);
+    left: -(@su1);
+
+    height: @su6;
```
dialpad-drew: 4px height

dialpad-drew (commented):
Screen Shot 2021-08-27 at 1 16 46 PM

Looks like the position change is creating a z-index collision. 

Also, because these are shims, I'm on the fence about documenting them. @dbecherdialpad do you have an opinion?

lib/build/less/components/toast.less
dbecherdialpad: Right. Don't use the utility classes as LESS mixins like this. As for CSS variables vs. LESS variables, I suppose I don't have as strong an opinion but we should follow the same standards as we are using elsewhere in dialtone.

dbecherdialpad:
@dialpad-drew If we don't plan on supporting these long-term then we should probably not document them. Maybe some comments in the LESS.

dialpad-drew:
> @dialpad-drew If we don't plan on supporting these long-term then we should probably not document them. Maybe some comments in the LESS.

In the long term, we don't plan to support them as a type of toast like they currently are. They're arguably a discrete component.

lib/build/less/components/toast.less
juliodialpad: Sorry, I copy-pasted the styles in uc_client to keep it the same, I'll update it

lib/build/less/components/toast.less
juliodialpad: Actually border-top is a lot better, I'll update it, thanks

juliodialpad (commented):
I added them to documentation for easier testing, should I remove it or just comment out lines in case we want to get it back?

dialpad-drew:
> I added them to documentation for easier testing, should I remove it or just comment out lines in case we want to get it back?

Let's remove them for now.

dialpad-drew:
I'm not sure whether this will become apparent on the product or not, but I suspect we're seeing this here because neither the table of contents nor the toast class with `position: fixed` has a define `z-index` value. So, by default, they're both `z-index: auto`. 

Toast _does_ have a defined `z-index` value, but its set on a child class, not the class that defines its position. 🤔

Screen Shot 2021-08-30 at 11 46 19 AM

juliodialpad:
> I'm not sure whether this will become apparent on the product or not, but I suspect we're seeing this here because neither the table of contents nor the toast class with `position: fixed` has a define `z-index` value. So, by default, they're both `z-index: auto`.
> 
> Toast _does_ have a defined `z-index` value, but its set on a child class, not the class that defines its position. 🤔
> 
> Screen Shot 2021-08-30 at 11 46 19 AM

I forgot to revert that, also I push the merged branch by accident I just wanted to push the updated CSS so you can review it. I'm trying to revert the merge and I'm going to revert the toast position to **absolute**

PR #422, opened 2021-08-31
Adds components top-level page

dbecherdialpad:
## Description
I found the current behaviour of just going directly to "Avatar" somewhat confusing, and I think it's worth calling out what is documented in Dialtone and that users should look at Dialtone-Vue.

## Pull Request Checklist

 - [ ] Provide a description what is being changed, extended, or removed.
 - [ ] Link to any issue(s) this request is resolving.

## Obligatory GIF
avatar


docs/components/index.html
```diff
@@ -1,8 +1,10 @@
...
 title: Components
-description: TBD.
+description: Browse Dialtone CSS components using the navigation on the left.
 ---
-<section class="dialtone--section">
-  TBD
+<section class="d-stack16">
+  {% header "h2", "Using Dialtone Components" %}
+  {% paragraph %}Dialtone documents the provided design system components, their intended usage, and relevant accessibility considerations for all implementations.{% endparagraph %}
+  {% paragraph %}Vue is the preferred method to use Dialtone components, as it's more robust, well-tested, and has these considerations built in. <a href="https://vue.dialpad.design/" class="d-link" target="_blank">Get started with Dialtone-Vue</a>.{% endparagraph %}
```
dbecherdialpad, line 9: This could be better. I am open to whatever makes sense here.

dbecherdialpad:
@dialpad-drew do you think this is worth doing?

dialpad-drew:
To be honest, I'm not entirely sure what we're solving for here. If it's site IA, then I think it should be done as a holistic IA enhancement, not just with the components (see Atlassian's design system as an example for top level nav items all have category pages).

If this is more about raising the visibility of Vue components vs. CSS components, I'm not sure I understand how this does that. 

Can you clarify for me the underlying problem?

dbecherdialpad:
it's solving for the fact that the Dialtone Vue homepage links here and it currently looks like this

Screen Shot 2021-11-19 at 11 07 42 AM

dialpad-drew:
> it's solving for the fact that the Dialtone Vue homepage links here and it currently looks like this
> 
> Screen Shot 2021-11-19 at 11 07 42 AM

Gotcha. Why not just change the link's URL to a page we don't intend people to visit at the moment? Seems like the cheapest fix for the moment.

In general, I think category pages are good, but it should be done across for similar site content (see Utilities), not just for one section.

dbecherdialpad:
> Why not just change the link's URL to a page we don't intend people to visit at the moment? Seems like the cheapest fix for the moment.

There's no page currently anywhere that specifically describes "dialtone components". The only option I suppose is to change the link to the dialtone homepage...then change it back when we add this category page later.

Not to be a pain, but are we letting perfect get in the way of good here? There's nothing stopping us from improving this later and doing the same for utilities and other sections. I only did this one because this is the only category we link to externally. Is there already a ticket for the category pages?

dialpad-drew:
> There's no page currently anywhere that specifically describes "dialtone components". The only option I suppose is to change the link to the dialtone homepage...then change it back when we add this category page later.

> Not to be a pain, but are we letting perfect get in the way of good here? There's nothing stopping us from improving this later and doing the same for utilities and other sections. I only did this one because this is the only category we link to externally. Is there already a ticket for the category pages?

It's all good, dude. You're not a pain -- I appreciate the opportunity to discuss this. 😄  "Perfect in the way of good" is always a fair question, but I don't think so. IMO, this is more "good getting in the way of superfluous" and here's why:

We currently have a site that, for better or worse, has a consistent navigational pattern for all top-level pages. That creates an expectation and confidence about how each of the site's sections functions. This proposal suggests breaking that navigational pattern and introducing a page with what appears to be mostly filler content, not because there's user feedback or we have a plan to gradually enhance the site's navigation overall or we have solid, useful content that needs to be communicated, but because a link on Storybook isn't ideal. I think that's the tail wagging the dog. 🐕 

In principle, I don't think we should be in a habit of yeeting pages or content onto the doc site. It may seem like a trivial thing but, in the aggregate, a bad content management practice can kill us. Fact is: no one reads -- [they muddle through](https://medium.com/@aniket.ambekar/lessons-learned-from-the-book-dont-make-me-think-by-steve-krug-8eddc339d213) -- and people are likely going to skim the site for very transactional, in-the-moment needs. The content we do offer should be exceptionally valuable, lean, and have a demonstrable raison d'etre. I'm not convinced this suggestion's rationale as I understand it meets that standard.

dbecherdialpad:
> ...what appears to be mostly filler content, not because there's user feedback...

I disagree. There is information about "dialtone components" as a general concept--what they are, what is a good component, what the difference is between Dialtone and Dialtone-Vue, when we write Dialtone CSS vs utility classes in Dialtone-Vue etc--that we don't communicate clearly anywhere else (that I know of). I thought we'd agreed on this at some point? If not here, where should we put that info? Part of my motivation for this was people ask and I don't have anywhere to direct them.

> Fact is: no one reads -- they muddle through...

I know. I have the original edition of that book on the shelf behind me 😉 

I am going to close this if there are other plans to improve the IA of the docs site. How would you suggest we link to Dialtone components from Dialtone-Vue? Just link to dialpad.design, or 302 redirect from `/components` to `/components/avatar`, or...?

PR #423, opened 2021-09-02
Patches issue with Archivo integration

dialpad-drew:
## Description
This PR un-ignores the `lib/build/fonts` folder and adds the necessary Archivo `woff2` files. It then updates the url in the `font-family` declaration.


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@dialpad/dialtone",
   "description": "The design system for Dialpad and UberConference",
-  "version": "5.23.0",
+  "version": "5.23.1-beta.1",
```
braddialpad: Do we still want to call this a beta version?

braddialpad (approved):

package.json
dialpad-drew: Good call.

PR #424, opened 2021-09-03
Updating deps

dialpad-drew:
## Description
Because dependabot is an obnoxious bot.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6681,9 +6681,9 @@
       }
     },
     "minizlib": {
-      "version": "2.1.0",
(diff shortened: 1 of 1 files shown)


dbecherdialpad (approved):
Can we just turn that off? There's really no risk of us somehow shipping security vulnerabilities with Dialtone since it's just CSS. The vast majority of these are ReDoS "attacks" which are inconsequential for dev dependencies.

dialpad-drew:
> Can we just turn that off? There's really no risk of us somehow shipping security vulnerabilities with Dialtone since it's just CSS. The vast majority of these are ReDoS "attacks" which are inconsequential for dev dependencies.

Yup. Done.

PR #427, opened 2021-09-09
button sizing fixes

braddialpad:
There are a number of places in dialpad where we have fairly complex components within a button. (ex/ header user info button, leftbar contact button). In these cases the icons within them were getting overidden by the button d-svg size even though they are very much not your "standard button icons". These are quite the pain to override in the product since the svg sizes in dialtone are !important. This change isolates the svg sizes to only apply if they are within a d-btn__icon, which I think makes a lot more sense anyway.


lib/build/less/components/button.less
```diff
@@ -74,7 +74,7 @@
         transition: none;
     }
 
-    .d-svg {
+    .d-btn__icon .d-svg {
         .d-svg--size16();
```
dbecherdialpad: maybe we could just (or also) remove the `!important` from `d-svg--sizeXX`? I have noticed a few potential issues with it and I don't think it's necessary for that class.

dbecherdialpad: I suppose we'd still need your fix based on the description haha

braddialpad:
@dialpad-drew do we need the !important? must be there for a reason right?

dbecherdialpad:
I thought it was only there bc it's a utility class and we add `!important` to all of them. So maybe we should follow our own advice and not use the utility classes as mixins lol

dialpad-drew:
I think David answered it. The utility classes are meant to be immutable.

braddialpad:
Removed the mixins

lib/build/less/components/button.less
```diff
@@ -74,8 +74,9 @@
         transition: none;
     }
 
-    .d-svg {
-        .d-svg--size16();
+    .d-btn__icon .d-svg {
+        width: 1.6rem;
+        height: 1.6rem;
```
ryandialpad: @braddialpad Maybe we should leverage the icon size variables here

```less
@icon12:                        1.2rem;
@icon14:                        1.4rem;
@icon16:                        1.6rem;
@icon18:                        1.8rem;
@icon20:                        2rem;
@icon24:                        2.4rem;
@icon32:                        3.2rem;
@icon48:                        4.8rem;
@icon64:                        6.4rem;
```

braddialpad: ah yep thanks

braddialpad: fixed

ryandialpad (approved):

lib/build/less/components/button.less
```diff
@@ -88,10 +89,10 @@
     padding: calc(var(--su6) - var(--su1)) calc(var(--su8) - var(--su1));  // Subtract border
     font-size: var(--fs12);
 
-    .d-svg {
+    .d-btn__icon .d-svg {
         // no icon-size 12 exists so setting manually
```
dbecherdialpad: we can remove this comment?

dbecherdialpad (approved):
for consistency, should we make those `iconX` variables available as CSS vars as well?

lib/build/less/components/button.less
braddialpad: done

braddialpad:
> for consistency, should we make those `iconX` variables available as CSS vars as well?

good call, added these.

dbecherdialpad:
excellent, then I can update the migration script to migrate the less mixins to css vars instead of hard-coded values.

PR #428, opened 2021-09-09
Remove transform from banner

braddialpad:
The transform method of positioning the banner does not work in dialpad due to having multiple banners that are not yet dialtone-vue banners. For now we need to render it at it's current position in the DOM

Since the dialtone documentation site uses the transform style, I have left it in as a separate class which is only used by the documentation site.


ryandialpad (approved):

docs/assets/css/dialtone.css
```diff
@@ -36750,6 +36758,7 @@ body.theme-dark {
     position: fixed !important;
   }
   .sm\:d-ps-sticky {
+    position: -webkit-sticky !important;
```
dbecherdialpad: do you need a dependency update or something? looks like these unnecessary vendor prefixes were re-added

dbecherdialpad (approved):

docs/assets/css/dialtone.css
braddialpad: hmm let me try

braddialpad: yep, that fixed it, thanks

PR #429, opened 2021-09-10
fix input icon misalignment

braddialpad:
If a parent element font size was set to anything other than 16, it would misalign the icon within the input. Fixed this by setting font-size to 16 on d-input-icon. Turns out font-size needs to be 16 regardless of icon size because the margin-top is calculated based on that.

Margin top calculations were also .1 off center, so fixed that as well.


lib/build/less/components/input.less
```diff
@@ -178,7 +178,8 @@
     position: absolute;
     z-index:  var(--zi-base1);
     margin: 0;
-    margin-top: 1.5rem;
+    margin-top: var(--su14);
+    font-size: @fs16;
```
dbecherdialpad: I am confused, `rem` is relative to the root element, not this element's font size. How does this fix the issue?

braddialpad:
> I am confused, `rem` is relative to the root element, not this element's font size. How does this fix the issue? 

What if it inherits 12px from the root because there is no font-size set?

dbecherdialpad:
`rem` is "root em", it's not inherited from the parent. everything in dialtone is predicated on the root font-size being reset to 10px

braddialpad:
Changed to line height to zero instead of setting font size. the smaller size margins needed to be adjusted but are still pretty "magic". Seems to work fine though.

dbecherdialpad (approved):

PR #437, opened 2021-09-17
add d-input-icon--right/left classes directly to the input

braddialpad:
d-input-icon--right and left classes will now have to be applied directly to the input as well as the icon to avoid selecting the input via the use of a "next sibling" selector. This was causing complicated issues with tab order in the product.

Dialtone-vue will also need an update for this to work properly. This will be coming shortly.


lib/build/less/components/input.less
```diff
@@ -188,26 +244,13 @@
...
-}
 
-//  $$ POSITIONS
-//  ----------------------------------------------------------------------------
-.d-input-icon--left {
-    left: var(--su8);
-
-    // Update padding for d-input, d-textarea if d-input-icon--left is present
-    & ~ .d-input,
-    & ~ .d-textarea {
```
dbecherdialpad: could we leave both for backward compatibility?

docs/components/input.html
```diff
@@ -242,7 +242,7 @@
                 </div>
                 <div>
                   <span class="d-input-icon d-input-icon--left">{% iconSystem "send" %}</span>
-                  <input class="d-input" id="Dialtone--InputExample--IconLeft" type="text" placeholder="Placeholder" />
+                  <input class="d-input d-input-icon--left" id="Dialtone--InputExample--IconLeft" type="text" placeholder="Placeholder" />
```
dbecherdialpad, line 245: also we should mention these classes in the docs

lib/build/less/components/input.less
braddialpad: hmm I think so?? let me check if anything breaks

braddialpad: added it back in... didn't seem to break anything but it's a lot of junk to leave in there

docs/components/input.html
braddialpad: yep, added

docs/_data/input.json
```diff
@@ -46,15 +46,25 @@
...
         "class": "d-input-icon--left",
         "applies": "d-input-icon",
-        "description": "Styles for left icons."
+        "description": "Styles for left icons. applied to the icon wrapper."
       },
       {
         "class": "d-input-icon--right",
         "applies": "d-input-icon",
-        "description": "Styles for right icons."
+        "description": "Styles for right icons. applied to the icon wrapper."
```
ryandialpad, line 67: @braddialpad Having these classes share the same name might be a bit confusing.

We should either indicate that one set's use is deprecated or use a different name for the new set, maybe using the term `with icon` instead of just input icon since it sounds like you are going to modify the icon element directly and not through the input element.

lib/build/less/components/input.less
```diff
@@ -98,15 +106,39 @@
 //  ----------------------------------------------------------------------------
 .d-input--xs {
     .input-button-xs();
+    &.d-input-icon--right {
+        padding-right: @inputs-icon-spacing--xs
```
ryandialpad: Nit: missing semi colons

docs/_data/input.json
ryandialpad: Looking at the comments in the build file, looks like we should mark the old classes as deprecated

docs/components/input.html
```diff
@@ -261,12 +261,12 @@
 <div>
   <label class="d-label" for="Dialtone--InputExample--IconLeft">...</label>
   <span class="d-input-icon d-input-icon--left">...</span>
-  <input class="d-input" id="Dialtone--InputExample--IconLeft" type="text" placeholder="..." />
+  <input class="d-input d-input-icon--left" id="Dialtone--InputExample--IconLeft" type="text" placeholder="..." />
```
ryandialpad, line 264: Looks like we have both the deprecated input icon modifier class and the new class being used here. We should probably remove the usage of the deprecated modifier classes.

docs/_data/input.json
braddialpad: there's no deprecated set though. It has the same name as before. You just have to put it on both the input and the icon now

lib/build/less/components/input.less
braddialpad: thanks, fixed

docs/_data/input.json
braddialpad: for reference the "deprecated" way is to use d-input-icon--right only on the icon and not the input

ryandialpad (approved):
w/ resolution of @dbecherdialpad 's discussion

dbecherdialpad (approved):

PR #438, opened 2021-09-17
Remove top and bottom margins from input components

ryandialpad:
## Description
Provide a description of what this pull request is adding, updating, extending, or removing. If this request is addressing an issue, please link to that issue.

Removing top and bottom margins from input components (including forms and selects). We will leave it up to the developer to correctly arrange our input, form and select components instead of assuming the order in which the elements will be defined.

This branch will likely need to be updated once @braddialpad's icon PR is merged (the input icons will currently be out of alignment).

After this is merged we will need to release and update Dialtone Vue, then release a new version of Dialtone Vue with the fix.

I also discovered that we might need to investigate and update our documentation for inputs and selects to be a bit more accessible / provide more accessible examples.

## Pull Request Checklist


lib/build/less/components/forms.less
```diff
@@ -122,11 +120,6 @@ fieldset {
   }
 }
 
-.d-checkbox__copy .d-validation-message,
-.d-radio__copy .d-validation-message {
-  margin-top: var(--su2);
-}
```
ryandialpad, line 128: `d-checkbox__copy` and `d-radio__copy` no longer exist

hynes-dialpad:
@ryandialpad One thought: The [Auto Spacing utilities](https://dialpad.design/utilities/spacing/auto-spacing/) might be good to use here instead of allowing engineers to specify margins. The only drawback is that this would uniformly set the top margin on all siblings (`* + *`), so if you want to reset or change the spacing you need to wrap elements within a new div.

Just a thought.

cc @dialpad-drew

ryandialpad:
> @ryandialpad One thought: The [Auto Spacing utilities](https://dialpad.design/utilities/spacing/auto-spacing/) might be good to use here instead of allowing engineers to specify margins. The only drawback is that this would uniformly set the top margin on all siblings (`* + *`), so if you want to reset or change the spacing you need to wrap elements within a new div.
> 
> Just a thought.
> 
> cc @dialpad-drew

I actually had that same thought! In the updated docs I added `d-stack6` and will be adding the utility class to the Dialtone Vue components.

dbecherdialpad:
Regardless of whether this is a good change to make or not, this could break a decent number of things. In fact a lot of our UC migration was around accounting for this new margin which we'd now need to go back and fix again.

ryandialpad:
> Regardless of whether this is a good change to make or not, this could break a decent number of things. In fact a lot of our UC migration was around accounting for this new margin which we'd now need to go back and fix again.

I saw this issues in a few places in UV. Wasn't aware we had manual migrations in UC to address this. My worry going forward is that we need a workaround every time we use an input or select without an accompanying label and / or description

dbecherdialpad:
> My worry going forward is that we need a workaround every time we use an input or select without an accompanying label and / or description

should we be doing that though?

ryandialpad:
> > My worry going forward is that we need a workaround every time we use an input or select without an accompanying label and / or description
> 
> should we be doing that though?

IMO, no however...

If we want to discourage this usage then we should probably update our Dialtone documentation which currently states the following

```
An input is normally paired with a label, but there are times when it can be used without a label. Placeholder text should primarily be used as a content prompt and only provided when needed.
```

and update Dialtone Vue so that a provided label is required.

dbecherdialpad:
> If we want to discourage this usage then we should probably update our Dialtone documentation which currently states the following...

well both are a valid use case (using aria-label or a visible label) but my point was if you're not using a visible label then you should probably know what you're doing.

@ryandialpad/@dialpad-drew Can I get a bit more clarity on where the issue is? What's changed between Dialtone 5 and 6 here? Is the issue just when there's no label or is it present with the label as well? There was a UC bug linked to this https://switchcomm.atlassian.net/browse/UC-11329

braddialpad:
Looking at the old inputs.less file from dialtone 5, I don't see any margin-top being set at all. It seems that the label spacing was set by line-height rather than margin

ryandialpad:
> Looking at the old inputs.less file from dialtone 5, I don't see any margin-top being set at all. It seems that the label spacing was set by line-height rather than margin

The issue occurs when there is no provided visible label.

The top margin was included in the design and we probably had good intentions initially, assuming that there wouldn't be cases where a visible label is not provided.

Looking back at the decision though, I believe for simplicity and flexibility that we should not be imposing this forced structure and leave the spacing up to the engineer using the component. In their case they would most likely be using the Dialtone Vue component which would have the correct implementation of the component (with the correct spacing classes added).

This decision however, could and probably will impact our current migration so that also needs to be taken into account as there might / will be additional effort required.

dbecherdialpad:
It feels like I've had to do a lot of digging to understand this issue haha. So the problem is how the `.d-input` (and related) elements now have a `margin-top` to space them from the label, which is new in Dialtone 6? And that doesn't make sense if there's no `.d-label`? If that's all it is, I think that's pretty simple and worthy of a fix.

Is it not as easy as just moving the margin to the bottom of `.d-label`? Or even something like:

```css
.d-label + .d-input {
  margin-top: ...
}
```

(this actually wouldn't work in `dialtone-vue` because of the dom structure wrapping the input in a wrapper element, but we could make it work my making that wrapper something like `.d-input__wrapper`. In fact looking at it, ~a lot of the util classes in the dialtone-vue component should probably be moved to dialtone itself~ EDIT ok it's actually just the `position: relative` on the wrapper)

> I believe for simplicity and flexibility that we should not be imposing this forced structure and leave the spacing up to the engineer using the component.

Not sure I understand this. We want to enforce consistent spacing between the label and input, no? If we leave this up to the engineer we'll end up with all sorts of weird spacing (more than we already have lol). Isn't that the point of dialtone?

dbecherdialpad:
And have you tried this change out locally against the UC codebase to see how many inputs are affected? I actually don't know.

ryandialpad:
> It feels like I've had to do a lot of digging to understand this issue haha. So the problem is how the `.d-input` (and related) elements now have a `margin-top` to space them from the label, which is new in Dialtone 6? And that doesn't make sense if there's no `.d-label`? If that's all it is, I think that's pretty simple and worthy of a fix.
> 

Correct, the issue is with automatically adding top margins to space label, description and the input element.


> Is it not as easy as just moving the margin to the bottom of `.d-label`? Or even something like:
> 
> ```css
> .d-label + .d-input {
>   margin-top: ...
> }
> ```
> 
> (this actually wouldn't work in `dialtone-vue` because of the dom structure wrapping the input in a wrapper element, but we could make it work my making that wrapper something like `.d-input__wrapper`. In fact looking at it, ~a lot of the util classes in the dialtone-vue component should probably be moved to dialtone itself~ EDIT ok it's actually just the `position: relative` on the wrapper)
> 
> > I believe for simplicity and flexibility that we should not be imposing this forced structure and leave the spacing up to the engineer using the component.
> 
> Not sure I understand this. We want to enforce consistent spacing between the label and input, no? If we leave this up to the engineer we'll end up with all sorts of weird spacing (more than we already have lol). Isn't that the point of dialtone?

I was going to say we should be encouraging engineers to just use the Vue components but after reflecting on the migration; you're entirely right, if we give an option to introduce inconsistent spacing there will be inconsistent spacing. We need to maintain a degree of opinion that is enforced with this design system.

Closing this PR. As a follow-up, should we update the Dialtone documentation to discourage the use of inputs without visible labels or warn that margins are automatically applied?

dialpad-drew:
I'm confused. If we want ~the~ to mandate spacing between inputs and labels, but the input also needs to hang on its own without extra margin, why don't we just use `margin-bottom` on the label like David said? 

Same for description text: why not just add the margin to the top of it? What am I missing here?

ryandialpad:
> I'm confused. If we want ~the~ to mandate spacing between inputs and labels, but the input also needs to hang on its own without extra margin, why don't we just use `margin-bottom` on the label like David said?
> 
> Same for description text: why not just add the margin to the top of it? What am I missing here?

We could do that. I think the original intention was to leave margins out of the label in case it is used on it's own.

dbecherdialpad:
> I think the original intention was to leave margins out of the label in case it is used on it's own.

That's valid. However, it's much less common than `.d-label` used on its own.

dialpad-drew:
I suspect it's more likely we'll have inputs of all kinds used without visible labels in specific and reasonable scenarios: checkboxes in tables, text input and text areas for chat/message composition, text input for search, etc. IMHO, I'd rather be in a position that margin has to be manually added to an a la carte `.d-label` than margin overridden and removed from the various input elements.

dbecherdialpad:
> I suspect it's more likely we'll have inputs of all kinds used without visible labels in specific and reasonable scenarios

Yep agreed. Looking through the code I can see that's already the case.

> IMHO, I'd rather be in a position that margin has to be manually added to an a la carte .d-label than margin overridden and removed from the various input elements.

I'm proposing that the easiest fix is to just add `margin-bottom` to `.d-label` (and `.d-description`) and remove margin-top as Ryan had here. In the _very rare_ case you're using `.d-label` in some other way you could just remove the bottom margin. I can see most of the manual usages of `.d-label` are already adding bottom margin to it.

@ryandialpad you closed this PR, but the discussion seems to still not be resolved. I think we'll want the change you made but with the addition of the margin bottom on the label. Does this seem like a good solution?

ryandialpad:
@dbecherdialpad @dialpad-drew If we are going this route, should we also add `margin-top` to validation messages? Rather than relying on developers adding `d-stack6`

Edit: Reverted change to validation message (top margin was removed in this PR)

ryandialpad:
Screen Shot 2021-09-28 at 6 09 48 PM
Screen Shot 2021-09-28 at 6 09 28 PM

Not entirely sure if this is what we are going for with the spacing between the label and description. I can look into modifying the selectors if this is not what we are looking for.

@dbecherdialpad 

Looking at the design, there shouldn't me any margin between the label and description. I will look into the selectors

ryandialpad:
Design: https://www.figma.com/file/zP2uyVxAg4notGNZwW64mW/Components-(6.0)?node-id=5515%3A408

docs/components/select.html
```diff
@@ -38,7 +38,7 @@
       {% codeWell %}
         {% codeWellHeader %}
           <div class="d-stack24 d-w100p">
-            <div>
+            <div class="d-stack6">
```
dbecherdialpad: I thought we were removing the need for this?

dbecherdialpad (changes requested):
It seems like I can't actually select any elements in that Figma to see what the margins should be. But now that we have a wrapper class, could something like this work?


```css
.d-label, .d-description {
  + .d-input, + .d-input__wrapper {
    margin-top: var(--su6);
  }
}
```

lib/build/less/components/forms.less
```diff
@@ -76,14 +78,19 @@ fieldset {
...
   color: var(--black-500);
   fill: currentColor;
   font-size: var(--fs12);
   font-family: inherit;
   line-height: var(--lh4);
 }
 
+.d-label ~ .d-description {
+  margin-top: var(--sun6);
+}
```
ryandialpad: @dbecherdialpad Might not be the best / simplest solution but here me out:

---

Firstly, there is no previous sibling selector

---

An alternative solution would be to only add the bottom margin if an input, select or textarea is present.

```css
.d-label ~ .d-input,
.d-label ~ .d-textarea,
.d-label ~ .d-select,
.d-description ~ .d-input,
.d-description ~ .d-textarea,
.d-description ~ .d-select {
  margin-top: var(--su6);
}
```

There are limitations to this:

If the label and / or description are not a sibling of the input / textarea / select then this selector will not apply.
If the "input" component is not an input / textarea / select then this selector will also not apply

---

Another alternative solution would be creating a label / description wrapper class. For simplicity, I wanted to avoid this solution.

```css
.d-some-wrapper-class {
  margin-bottom: var(--su6);
}
```

```html
<div class="d-some-wrapper-class">
  <div class="d-label">...</div>
  <div class="d-description">...</div>
</div>
<div class="d-input__wrapper">
  ...
  <input class="d-input">...</input>
  ...
</div>
...
```

docs/components/select.html
ryandialpad: We are, missed that one, still going through the changes

lib/build/less/components/forms.less
dbecherdialpad: I am not super worried about supporting the generic case, I think it's reasonable to assume that label + description + input are all adjacent siblings, but this discussion has gone on quite a bit so I am fine with the negative margin approach.

For that I think this should be: `.d-label + .d-description`

ryandialpad:
> It seems like I can't actually select any elements in that Figma to see what the margins should be. But now that we have a wrapper class, could something like this work?
> 
> ```css
> .d-label, .d-description {
>   + .d-input, + .d-input__wrapper {
>     margin-top: var(--su6);
>   }
> }
> ```

I was actually writing up the three potential solutions that I could think of, thoughts @dbecherdialpad?

Using the input wrapper class, we might want to create a label / description wrapper class as well based on the provided code snippets

lib/build/less/components/forms.less
ryandialpad: Pushed changes, verifying locally that everything still looks good

dbecherdialpad:
A wrapper seems unnecessary. This is actually a very simple problem. The adjacent sibling selector seems sufficient for what we need.

lib/build/less/components/input.less
```diff
@@ -244,7 +243,7 @@
     top: 0;
     z-index:  var(--zi-base1);
     margin: 0;
-    margin-top: 1.6rem;
+    margin-top: 1rem;
```
ryandialpad, line 246: Need to reduce icon top margin by 0.6rem as we removed the input top margin by 0.6rem

docs/components/select.html
dbecherdialpad: ok so...are we? do we need this or not?

ryandialpad: We don't, removing them, thought there was just one.....boy was I wrong

docs/components/select.html
```diff
@@ -121,10 +123,14 @@
         {% endcodeWellHeader %}
         {% codeWellFooter %}
           {% highlight html linenos %}
-<div class="d-label">...</div>
-<div class="d-description">...</div>
-<div class="d-select">
-  <select class="d-select__input" id="Dialtone--SelectExample3">...</select>
+<div class="d-w100p">
+  <div>
```
dbecherdialpad: what's up with the extra `<div>`?

dbecherdialpad (approved):
boy howdy, we got there eventually. thanks!

docs/components/select.html
dbecherdialpad: u fixed this as I was reviewing I think. couldn't delete the comment

PR #439, opened 2021-09-20
Added display: block to svg to help center the elements

juliodialpad:
## Description
The action icons present in the conversation window are not center aligned.

I think that many SVG icons are misaligned because of the lack of display: block.

https://switchcomm.atlassian.net/browse/DP-40569

## Pull Request Checklist


lib/build/less/dialtone-globals.less
```diff
@@ -42,6 +42,7 @@ body {
     //  By default, most icons are ~24px.
     width: 1.5em;
     height: 1.5em;
+    display: block;
```
dbecherdialpad: I thought we determined this was probably because of `line-height`?

juliodialpad: No, seems like there was a confusion. 

line-height helps to align better some items, but doesn't solve all the problems.

While display: block; aligns a lot if not all the elements that are misaligned

dbecherdialpad: I am still a little iffy on this, we shouldn't need to override the default display property of all SVGs just to align them. There's no real reason I can think of that the default display value would cause misalignment other than `line-height`. If I add `line-height: 0` to `d-btn__icon` it fixes the issue for this ticket, is that not always the case? I suppose it doesn't help that we use the icons in about one thousand slightly different ways.

Screen Shot 2021-09-20 at 4 40 02 PM

Also keep in mind that adding `display:block` mean that by default, without a size class, they will now be 100% width

juliodialpad: Seems fair, I'll add line-height: 0 to .d-btn__icon instead

dbecherdialpad (approved):

PR #440, opened 2021-09-20
DT-103: Fix icon button margins

ryandialpad:
## Description
Bugfix for icon button margins:

We now only add margins if the icons are not the only child of button.
Also added fixes to order to ensure that the icons are correctly ordered no matter how they are defined in the DOM.

## Pull Request Checklist

Screen Shot 2021-09-20 at 11 46 16 AM

### DIFF
--- a/lib/build/less/components/button.less
+++ b/lib/build/less/components/button.less
@@ -167,13 +167,13 @@
     margin-right: var(--sun2);
 }
 
-.d-btn__icon--left + .d-btn__label,
-.d-btn__label + .d-btn__icon--right {
+.d-btn__icon--left ~ .d-btn__label,
+.d-btn__label ~ .d-btn__icon--right {
     margin-left: var(--su4);
 }
 
-.d-btn__icon--right + .d-btn__label,
-.d-btn__label + .d-btn__icon--left {
+.d-btn__icon--right ~ .d-btn__label,
+.d-btn__label ~ .d-btn__icon--left {
     margin-right: var(--su4);
 }
 
--- a/lib/build/less/components/button.less
+++ b/lib/build/less/components/button.less
@@ -152,14 +152,29 @@
 }
 
 .d-btn__icon--left {
-    margin-left: var(--sun2);
-    margin-right: var(--su4);
+    order: -1;
 }
 
 .d-btn__icon--right {
     order: 1;
-    margin-left: var(--su4);
+}
+
+.d-btn__icon--left:not(:only-child) {
+    margin-left: var(--sun2);
+}
+
+.d-btn__icon--right:not(:only-child) {
     margin-right: var(--sun2);
+}
+
+.d-btn__icon--left + .d-btn__label,
+.d-btn__label + .d-btn__icon--right {
+    margin-left: var(--su4);
+}
+
+.d-btn__icon--right + .d-btn__label,
+.d-btn__label + .d-btn__icon--left {
+    margin-right: var(--su4);
 }
 
 //  ============================================================================


lib/build/less/components/button.less
```diff
@@ -152,16 +152,31 @@
 }
 
 .d-btn__icon--left {
-    margin-left: var(--sun2);
-    margin-right: var(--su4);
+    order: -1;
```
ryandialpad, line 155: Ensure that left icon is always on the left, no matter it's place in the dom relative to the label

lib/build/less/components/button.less
```diff
@@ -152,16 +152,31 @@
...
 }
 
 .d-btn__icon--right {
     order: 1;
```
ryandialpad, line 159: Ensure that right icon is always on the right, no matter it's place in the dom relative to the label

lib/build/less/components/button.less
```diff
@@ -152,16 +152,31 @@
...
-    margin-left: var(--su4);
+}
+
+.d-btn__icon--left:not(:only-child) {
+    margin-left: var(--sun2);
+}
+
+.d-btn__icon--right:not(:only-child) {
     margin-right: var(--sun2);
 }
```
ryandialpad, line 168: Only add exterior negative margins if the icon is not an only child. This assumes that we don't have buttons defined with two icons and no labels.

This was the cleanest approach I could think of as there is no previous sibling selector in the case that icons are added after the label.

lib/build/less/components/button.less
```diff
@@ -152,16 +152,31 @@
...
 
+.d-btn__icon--left + .d-btn__label,
+.d-btn__label + .d-btn__icon--right {
+    margin-left: var(--su4);
+}
+
+.d-btn__icon--right + .d-btn__label,
+.d-btn__label + .d-btn__icon--left {
+    margin-right: var(--su4);
+}
```
ryandialpad: Add inner margins between icons and label if a label is present

dbecherdialpad: won't this depend on the DOM order in contrast to the `order` change you made above?

dbecherdialpad (approved):

lib/build/less/components/button.less
ryandialpad: @dbecherdialpad The margins need to be added relative to where the label is located in respect to the icon. As such, I have added styles for all icons and positions which cover the following use cases:

* left icon previous sibling of label
* left icon next sibling of label
* right icon previous sibling of label
* right icon next sibling of label

ryandialpad: The following use cases need to be tested (although idk if we have any buttons with two icons use cases)

* both icons as next siblings of label
* both icons as previous siblings of label

ryandialpad: Update on testing the dual icon tests failed due the use of an immediate adjacent sibling selector instead of a general adjacent sibling selector. Will update to fix, my bad

ryandialpad: Pushed changes, will upload screenshot shortly

ryandialpad: Screen Shot 2021-09-20 at 1 04 17 PM

lib/build/less/components/button.less
braddialpad: just wanted to add for reference, that the order property does not affect tab order, only visual order.

dbecherdialpad: well these are in buttons already, but does it affect the order in which a screen reader reads them? I think no

braddialpad: nope also does not affect screen reader

Issue #330, opened 2021-05-26
Height/Width utility documentation is confusing


dbecherdialpad:
The height and width documentation has sections for percentage/fixed/etc heights, but they don't include all the options:

<img width="782" alt="Screen Shot 2021-05-26 at 11 45 03 AM" src="https://user-images.githubusercontent.com/42597813/119714557-db6fe200-be17-11eb-9de2-cf3030d26de7.png">

All the class options _are_ listed in the top scrolling table but it's not obvious, especially for any that are below the fold.

(This might affect other documentation sections as well)

PR #446, opened 2021-09-22
Updated width and height documentation

juliodialpad:
## Description
Height and width utility class documentation doesn't show examples for every available height and width class.
https://switchcomm.atlassian.net/browse/DT-16

Added height and width examples, changed the way the elements look, to avoid visual issues with small examples

## Pull Request Checklist


dbecherdialpad (dismissed):
Docs look off to me: "3" is actually 25% and then from there the numbers are off.
Screen Shot 2021-09-22 at 3 06 35 PM

Similarly for fixed heights, they're off by one (the examples start at 0):

Screen Shot 2021-09-22 at 3 07 04 PM

docs/utilities/sizing/height.html
```diff
@@ -53,19 +53,83 @@
...
+          </div>
+          <div class="d-d-flex d-fd-column d-pls-start d-h100p d-w64"><span>11</span>
+            <div class="d-h216">
+              <div class="d-h90p d-bgc-green-300 d-bar4"></div>
+            </div>
+          </div>
+          <div class="d-d-flex d-fd-column d-pls-start d-h100p d-w64"><span>12</span>
+            <div class="d-h216">
+              <div class="d-h100p d-bgc-green-300 d-bar4"></div>
+            </div>
```
ryandialpad: Can we clean this up using a for loop to iterate through the different height and width classes?

juliodialpad:
> Docs look off to me: "3" is actually 25% and then from there the numbers are off.
> Screen Shot 2021-09-22 at 3 06 35 PM
> 
> Similarly for fixed heights, they're off by one (the examples start at 0):
> 
> Screen Shot 2021-09-22 at 3 07 04 PM

The 3 above the bar means that the example "3" here:

```html
<div class="d-h216">
    <div class="d-h10p">1</div>
    <div class="d-h20p">2</div>
    <div class="d-h25p">3</div>
    <div class="d-h30p">4</div>
    <div class="d-h40p">5</div>
    <div class="d-h50p">6</div>
    <div class="d-h60p">7</div>
    <div class="d-h70p">8</div>
    <div class="d-h75p">9</div>
    <div class="d-h80p">10</div>
    <div class="d-h90p">11</div>
    <div class="d-h100p">12</div>
</div>
```

Gives you a height of 25%, not meaning that specific bar is 30% height.

Or should I put the percentage above the example?

docs/utilities/sizing/height.html
juliodialpad: Sure, I didn't knew that I could use for loops here

dbecherdialpad:
The description of this section says:

> Use `d-h{n}p` to set a percentage height for an element.

So I read the numbers to be the number you'd use in the class name. So I guess they should be 10, 20, 25, etc? Or am I reading it wrong?

juliodialpad:
The original documentation had numbers to illustrate the examples, that's why I was leaving as it was.

Captura de Pantalla 2021-09-23 a la(s) 1 31 59 p m

Just put the numbers above the bars because in the fixed examples, the first heights were to low to fit numbers inside them.

Captura de Pantalla 2021-09-21 a la(s) 3 07 36 p m

dialpad-drew:
> The description of this section says:
> 
> > Use `d-h{n}p` to set a percentage height for an element.
> 
> So I read the numbers to be the number you'd use in the class name. So I guess they should be 10, 20, 25, etc? Or am I reading it wrong?

You may be reading it wrong. It looks like Julio is following how some of the other documentation does this.

dbecherdialpad:
It just seems confusing now that we're showing all the height classes instead of an example subset. Why have 2 different sets of numbers?

dialpad-drew:
Well, there's also precedent for us showing the class with the example if that's the route we want to take: https://dialpad.design/utilities/spacing/padding/

dialpad-drew:
Based on what I'm seeing on other documentation, our options seem to be this:

1. Keep as-is and use numerical annotations to show the relationship between the mark-up and the visualization
2. Replace the numbers with the classes (`d-h0`, `d-h1`, etc.) so the class and the visualization match

I don't have an opinion other than this: when you step back, how much of an actual big deal is this @dbecherdialpad ? My hypothesis is someone will come to this page to see the classes, grab the one they need after scanning the page and without really reading/grokking the content, and close the page. Just my theory.

dbecherdialpad:
I don't think it's a big deal. TBH I don't think people really need an example of every single height class.

> My hypothesis is someone will come to this page to see the classes, grab the one they need after scanning the page and without really reading/grokking the content, and close the page.

I agree which is the only reason I bring it up, I think it would be easter to grok if the DT classes themselves were more prominent rather than having to match up the numbers with the code block.

dialpad-drew:
> I don't think it's a big deal. TBH I don't think people really need an example of every single height class.
> 
> > My hypothesis is someone will come to this page to see the classes, grab the one they need after scanning the page and without really reading/grokking the content, and close the page.
> 
> I agree which is the only reason I bring it up, I think it would be easter to grok if the DT classes themselves were more prominent rather than having to match up the numbers with the code block.

I'm confused. I thought part of the reason we're doing this is because you had a problem not listing all the class options: https://github.com/dialpad/dialtone/issues/330

dbecherdialpad:
I literally didn't even remember creating that issue, but as I recall the problem was more about the table at the top, and how it was not super obvious you had to scroll to see the percent based ones. We've since fixed that issue and it looks like we moved the percentage based classes to the top.

Now that I see it again though, it does make some sense to have examples for all the percentage options at least, I just personally think it's more clear to have the class name in the examples rather than numbers.

dialpad-drew:
> I literally didn't even remember creating that issue, but as I recall the problem was more about the table at the top, and how it was not super obvious you had to scroll to see the percent based ones. We've since fixed that issue and it looks like we moved the percentage based classes to the top.
> 
> Now that I see it again though, it does make some sense to have examples for all the percentage options at least, I just personally think it's more clear to have the class name in the examples rather than numbers.

Cool. Let's try it. 

CC: @juliodialpad For reference, here's a place we show and style the class in the example: https://dialpad.design/utilities/spacing/padding/

juliodialpad:
Looks much better to me this way.

Captura de Pantalla 2021-09-28 a la(s) 11 53 26 a m

dialpad-drew:
Can we style the percentage class names? Look at the mono font style and weight in the [padding documentation](https://deploy-preview-446--dialtone.netlify.app/utilities/spacing/padding/).

Should we include the class name in all the visualizations on this page rather than just the first one?

@dbecherdialpad is this heading in a better direction fro you?

dbecherdialpad:
yep this looks great

juliodialpad:
Any remaining comments here or can be approved to merge? @dbecherdialpad @dialpad-drew

dialpad-drew (approved):

dbecherdialpad:
> Should we include the class name in all the visualizations on this page rather than just the first one?

I think that's a good improvement we could make later. Let's merge this one.

PR #447, opened 2021-09-23
Github actions deploy workflow

braddialpad:
add github actions workflow to automatically build and deploy when the production branch is pushed.

It will also trigger on a branch named alpha or beta, and will automatically tag the deployment as the respective alpha or beta release.

Created a small release bash script to make it simpler to deploy with a single command, handles pushing the commits and tags at the same time

Netlify build commands updated.

All files generated by the build have been removed from the repo and added to gitignore.

See RELEASING.md for the new deployment process.


ryandialpad (approved):
LGTM!

Just to be 100% sure, do the fonts get copied during build to the dist folder?

braddialpad:
> LGTM!
> 
> Just to be 100% sure, do the fonts get copied during build to the dist folder?

Great catch they were NOT. Updated the gulpfile to output them to the dist folder

.github/RELEASING.md
```diff
@@ -1,30 +1,17 @@
...
-3. Run `git push` to push updated `package.json` to remote.
-4. Run `git push --tags`. This creates a `git` version tag in the project.
-5. Run `npm publish --tag prelease` For instance, `npm publish --tag beta`. It can then be installed in another project using the specific version number or the tag (i.e. `npm i packagename@beta`).
+1. Make sure your staging and production branches are up-to-date locally. You should be in the `production` branch. Also it’s a good idea to stop your local server while versioning.
+> Note if you would like to deploy a prerelease (alpha or beta). You would do this on a branch named `alpha` or `beta` instead of `production`. This will automatically tag the deployment to npm as a alpha or beta release so npm does not consider it to be the latest version. It could be installed with the command `npm i dialtone@beta`
+2. `git merge staging` to merge staging into the production branch to cut a new release.
+3. If bumping Dialtone-Vue version, update the badge manually in `/docs/index.html` and commit. (this will be automated soon)
+4. In your CLI window, run `./release.sh` from the dialtone repository directory.
```
dbecherdialpad: I feel like the version bump should happen on staging(?) so that the main branch always has the up-to-date version?

release.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# exit when any command fails
+set -e
+
+echo "Enter the version number of this release Ex/ 6.0.2 or 6.0.0-beta.16: "
```
dbecherdialpad: I think we should encourage just using `patch`/`minor`/`major` to this command rather than specifying an exact version number

.github/RELEASING.md
braddialpad: yeah we can do that

release.sh
braddialpad: updated this

.github/RELEASING.md
braddialpad: changed this doc to instruct pushing staging first, then merging to production

.github/RELEASING.md
```diff
@@ -1,30 +1,19 @@
...
+4. It will prompt you to enter a version. enter major, minor, or patch depending on what your change consists of. Major versions are breaking changes. Minor versions are large changes but backward compatible. Patches are bug fixes to existing Dialtone items. You may also enter an exact version if you need to.
+> Note on a prerelease our convention is major.minor.patch-prerelease.buildmetadata ex. `6.0.0-alpha.1`.
+6. After entering your version number the package.json version numbers will be updated and committed with a git tag matching the version you entered. This will then be automatically pushed to the remote
+5. `git checkout production` to move to the production branch
+> Note if you would like to deploy a prerelease (alpha or beta). You would do this on a branch named `alpha` or `beta` instead of `production`. This will automatically tag the deployment to npm as a alpha or beta release so npm does not consider it to be the latest version. It could be installed with the command `npm i dialtone@beta`
+6. `git merge staging` (or your prerelease working branch) to merge the changes to this branch
```
dbecherdialpad: should we use `rebase` (or maybe just `--ff-only`) instead of a merge commit since now the production branch should not really have commits not on staging?

.github/RELEASING.md
```diff
@@ -1,30 +1,19 @@
...
+7. `git push` to push the branch. This will trigger the deploy.
+7. You should be able to see your deploy running at https://github.com/dialpad/dialtone/actions
+8. When this has completed the new version of the package has been deployed to npm, and the documentation website (`dialpad.design`) is now updated.
+9. Merge the production branch back to staging so they remain in sync.
```
dbecherdialpad: this _shouldn't_ be necessary since we aren't really committing to production directly right?

.github/RELEASING.md
```diff
@@ -1,30 +1,19 @@
...
+10. Finally you need [draft a release on Github](https://github.com/dialpad/dialtone/releases/new). You can look at [past releases](https://github.com/dialpad/dialtone/releases/tag/v5.13.0) for reference, but generally I give it a title (What’s the update?) and then a short description of what was added, changed, or fixed.
```
dbecherdialpad: should we add a template for this?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: parse branch from ref
+        id: branch
+        run: echo ::set-output name=BRANCH::${GITHUB_REF/refs\/heads\//}
+
+      # Will prevent the rest of the steps from running on fail
+      - name: Check if user is a dialpad member
+        uses: octokit/request-action@v2.1.0
+        with:
+          route: GET /orgs/{org}/members/{username}
```
dbecherdialpad, line 32: is "dialpad member" sufficient or should we check membership of the Dialtone maintainers team?

braddialpad: We probably don't really even need this at all since I verified our branch protection rules on github (only admins/maintainers can push the branch). Thought I would just leave it there as an extra safeguard since I already wrote it

.github/RELEASING.md
braddialpad: Seems like you can't add a release template the same way you can for PRs/Issues, but I could just add an md file to reference. Looks like Drew already has some standardized headers so i'll copy that

.github/workflows/deploy.yml
dbecherdialpad: sure sounds good. the branch protection rule is better

.github/RELEASING.md
braddialpad: should be gone in recent update

.github/RELEASING.md
braddialpad: hmmm. Yeah I think that's better

.github/RELEASING.md
braddialpad: release template has been added.

.github/RELEASING.md
braddialpad: opted for the --ff-only on this one. Added to doc.

dbecherdialpad (approved):

PR #450, opened 2021-09-27
Immutable input validation states

josedialpad:
## Description
In DT5 we have the [border color of input validation states immutable](https://github.com/dialpad/dialtone/blob/v5.24.0/lib/build/less/components/inputs.less#L102). I'm updating to preserve it in DT6 as it will fix some use cases in Dialpad where they were overwriting input border color and the validation state border color wasn't applying because of the CSS specificity.

## Pull Request Checklist

 - [ ] Link to any issue(s) this request is resolving.

### DIFF
--- a/lib/build/less/components/input.less
+++ b/lib/build/less/components/input.less
@@ -80,7 +80,7 @@
     &:focus {
         --input--bc: var(--input--focus-bc);
 
-        box-shadow: var(--bs-focus-ring);
+        box-shadow: var(--bs-focus-ring)!important;
         outline: 0;
     }
     //  --  DISABLED / READ-ONLY
@@ -89,7 +89,7 @@
         --input--bc:           var(--black-500);
         --input--bgc:          var(--black-200);
 
-        box-shadow: none;
+        box-shadow: none!important;
 
         //  --  Placeholder copy
         &::placeholder {
@@ -201,7 +201,7 @@
     --input--focus-bc:  var(--warning-color);
 
     &:focus {
-        box-shadow: var(--bs-focus-ring-warning);
+        box-shadow: var(--bs-focus-ring-warning)!important;
     }
 }
 .d-input--error,
@@ -210,7 +210,7 @@
     --input--focus-bc:  var(--error-color);
 
     &:focus {
-        box-shadow: var(--bs-focus-ring-error);
+        box-shadow: var(--bs-focus-ring-error)!important;
     }
 }
 .d-input--success,
@@ -219,7 +219,7 @@
     --input--focus-bc:  var(--success-color);
 
     &:focus {
-        box-shadow: var(--bs-focus-ring-success);
+        box-shadow: var(--bs-focus-ring-success)!important;
     }
 }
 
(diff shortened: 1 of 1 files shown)


dialpad-drew (approved):

ryandialpad:
@josedialpad What happens if the input is disabled with these changes?

If we are making these variant classes immutable, we should also update their focus styles to be immutable as well (not just the border color).

josedialpad:
> @josedialpad What happens if the input is disabled with these changes?
> 
> If we are making these variant classes immutable, we should also update their focus styles to be immutable as well (not just the border color).

It should work as in DT5, overriding the disabled border color:

2021-09-27 at 18 36 32@2x

Maybe this is not desired and should have the disabled border color, so updated this in this PR https://github.com/dialpad/dialtone/pull/450/commits/7320d7ac945b3185cd312650eb70fdd9d4dda082

> we should also update their focus styles to be immutable as well (not just the border color).

I agree, though we didn't have this in DT5. Updated in DT6: https://github.com/dialpad/dialtone/pull/450/commits/0d07b3dff79735cef923c7847355298b8142c817

ryandialpad:
> > @josedialpad What happens if the input is disabled with these changes?
> > If we are making these variant classes immutable, we should also update their focus styles to be immutable as well (not just the border color).
> 
> It should work as in DT5, overriding the disabled border color:
> 
> 2021-09-27 at 18 36 32@2x
> 
> Maybe this is not desired and should have the disabled border color, so updated this in this PR [7320d7a](https://github.com/dialpad/dialtone/commit/7320d7ac945b3185cd312650eb70fdd9d4dda082)
> 
> > we should also update their focus styles to be immutable as well (not just the border color).
> 
> I agree, though we didn't have this in DT5. Updated in DT6: [0d07b3d](https://github.com/dialpad/dialtone/commit/0d07b3dff79735cef923c7847355298b8142c817)

This is a bit confusing as a user as that field doesn't 100% look disabled with the warning coloured outline.

An approach to get around this issue would be relying on a disable class which is applied when the element is disabled and contains immutable styles that cannot be overridden by the validation classes. This would require the disable class to be applied prior to the validation class since they would both be immutable.

Thoughts @dialpad-drew?

dialpad-drew:
> An approach to get around this issue would be relying on a disable class which is applied when the element is disabled and contains immutable styles that cannot be overridden by the validation classes. This would require the disable class to be applied prior to the validation class since they would both be immutable.

Why would we need to add another class? Why can't we handle this in the cascade?

josedialpad:
> > An approach to get around this issue would be relying on a disable class which is applied when the element is disabled and contains immutable styles that cannot be overridden by the validation classes. This would require the disable class to be applied prior to the validation class since they would both be immutable.
> 
> Why would we need to add another class? Why can't we handle this in the cascade?

Now it's working with the specificity:

2021-09-28 at 09 29 52@2x

ryandialpad:
> > An approach to get around this issue would be relying on a disable class which is applied when the element is disabled and contains immutable styles that cannot be overridden by the validation classes. This would require the disable class to be applied prior to the validation class since they would both be immutable.
> 
> Why would we need to add another class? Why can't we handle this in the cascade?

Good catch, that would be preferable. For so reason I though I remembered us discussing adding disabled classes to some components for some reason.

ryandialpad (approved):

PR #451, opened 2021-09-27
DT-Vue current version

braddialpad:
GHA will read the version number from dialtone-vue's package.json on deploy, and write it to the index.html file.


dbecherdialpad (approved):
ha, interesting. I suppose this is a lot easier if dialtone-vue were just public? And I suppose we can't trigger this automatically when dt-vue is published?

dialpad-drew:
@braddialpad do we need to update any release documentation to include this?

braddialpad:
> ha, interesting. I suppose this is a lot easier if dialtone-vue were just public?

Yep, much

> And I suppose we can't trigger this automatically when dt-vue is published?

hmm now that you mention it, we may HAVE to do it this way.. I just realized that if we deploy a dialtone version, and then dialtone-vue updates to a new version, we're still going to have the old version in dialtone without doing a redeploy..

braddialpad:
> @braddialpad do we need to update any release documentation to include this?

yeah thanks for the reminder, there is one line that will have to be removed about manually updating the version number

ryandialpad (approved):
Great work!

braddialpad:
Scrapping this PR as it will not show the correct version number if dialtone-vue updates without a dialtone update.

New solution can be found here: https://github.com/dialpad/dialtone/pull/453

PR #460, opened 2021-09-29
Add Checkbox 'Indeterminate' State styling

modialpad:
## Description

Add Checkbox 'Indeterminate' state styles. 

How it looks:

image

## Pull Request Checklist

### DIFF
--- a/docs/components/checkbox.html
+++ b/docs/components/checkbox.html
@@ -70,6 +70,30 @@ storybook-url: https://vue.dialpad.design/?path=/story/forms-checkbox--default
                   <label class="d-checkbox__label" for="Dialtone-CheckExample4">Checkbox label that's been disabled</label>
                 </div>
               </div>
+              <div class="d-checkbox-group">
+                <div class="d-checkbox__input">
+                  <input class="d-checkbox d-checkbox--indeterminate" type="checkbox" name="Dialtone-CheckExample5" id="Dialtone-CheckExample5" />
+                </div>
+                <div class="d-checkbox__copy">
+                  <label class="d-checkbox__label" for="Dialtone-CheckExample5">Checkbox that's in Indeterminate state with Custom Class</label>
+                </div>
+              </div>
+              <div class="d-checkbox-group">
+                <div class="d-checkbox__input">
+                  <input class="d-checkbox" type="checkbox" name="Dialtone-CheckExample6" id="Dialtone-CheckExample6" indeterminate />
+                </div>
+                <div class="d-checkbox__copy">
+                  <label class="d-checkbox__label" for="Dialtone-CheckExample6">Checkbox that's in Indeterminate state with 'indeterminate' attribute</label>
+                </div>
+              </div>
+              <div class="d-checkbox-group">
+                <div class="d-checkbox__input">
+                  <input class="d-checkbox d-checkbox--indeterminate" type="checkbox" name="Dialtone-CheckExample7" id="Dialtone-CheckExample7" disabled />
+                </div>
+                <div class="d-checkbox__copy">
+                  <label class="d-checkbox__label" for="Dialtone-CheckExample7">Checkbox that's in Indeterminate state and disabled</label>
+                </div>
+              </div>
           </fieldset>
         {% endcodeWellHeader %}
         {% codeWellFooter %}
--- a/docs/components/checkbox.html
+++ b/docs/components/checkbox.html
@@ -38,123 +38,138 @@ storybook-url: https://vue.dialpad.design/?path=/story/forms-checkbox--default
       {% codeWell %}
         {% codeWellHeader %}
           <fieldset class="d-stack8">
-              <div class="d-checkbox-group">
-                <div class="d-checkbox__input">
-                  <input class="d-checkbox" type="checkbox" name="Dialtone-CheckExample1" id="Dialtone-CheckExample1" />
-                </div>
-                <div class="d-checkbox__copy">
-                  <label class="d-checkbox__label" for="Dialtone-CheckExample1">Checkbox label</label>
-                </div>
+            <div class="d-checkbox-group">
+              <div class="d-checkbox__input">
+                <input class="d-checkbox" type="checkbox" name="Dialtone-CheckExample1" id="Dialtone-CheckExample1" />
               </div>
-              <div class="d-checkbox-group">
-                <div class="d-checkbox__input">
-                  <input class="d-checkbox" type="checkbox" name="Dialtone-CheckExample2" id="Dialtone-CheckExample2" checked />
-                </div>
-                <div class="d-checkbox__copy">
-                  <label class="d-checkbox__label" for="Dialtone-CheckExample2">Checkbox label</label>
-                </div>
+              <div class="d-checkbox__copy">
+                <label class="d-checkbox__label" for="Dialtone-CheckExample1">Checkbox label</label>
               </div>
-              <div class="d-checkbox-group d-checkbox-group--disabled">
-                <div class="d-checkbox__input">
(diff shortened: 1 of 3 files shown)


dialpad-drew:
Awesome, @modialpad thanks for PRing this. Glad to see you updated the documentation.

I just pushed a quick update as well; my suggestion is we parse the indeterminate style out to its own discrete section. LMK what you think.

Also, out of curiosity: is your local build looping when you save?

modialpad:
> Awesome, @modialpad thanks for PRing this. Glad to see you updated the documentation.
> 
> I just pushed a quick update as well; my suggestion is we parse the indeterminate style out to its own discrete section. LMK what you think.
> 
> Also, out of curiosity: is your local build looping when you save?

Np, looks more organized now 👍 

Yes, when I run `npm run start` I think it rebuilds everything whenever I edit any files. But I think there's a bug where if anything changes the doc server doesn't restart properly. I had to exist the process and re-run `npm run start` in order to see app loading properly on `localhost:4000`. Is it just me?

dialpad-drew:
> > Awesome, @modialpad thanks for PRing this. Glad to see you updated the documentation.
> > I just pushed a quick update as well; my suggestion is we parse the indeterminate style out to its own discrete section. LMK what you think.
> > Also, out of curiosity: is your local build looping when you save?
> 
> Np, looks more organized now 👍
> 
> Yes, when I run `npm run start` I think it rebuilds everything whenever I edit any files. But I think there's a bug where if anything changes the doc server doesn't restart properly. I had to exist the process and re-run `npm run start` in order to see app loading properly on `localhost:4000`. Is it just me?

No, that's not just you. I'm noticing that on my end as well; that's new. I already noted it.

Only question I have is about the attribute. [MDN suggests](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox) that browsers don't currently support `indeterminate` as an attribute and the example using the attribute doesn't seem to exhibit any functionality. Is there a CSS-only way to do this, or are we simply just putting in some styles for the Vue component to leverage?

docs/components/checkbox.html
```diff
@@ -38,75 +38,138 @@
...
+    <div class="d-checkbox__input">
+      <input class="d-checkbox d-checkbox--indeterminate" type="checkbox" name="Dialtone-CheckExample5" id="Dialtone-CheckExample5" />
+    </div>
+    <div class="d-checkbox__copy">
+      <label class="d-checkbox__label" for="Dialtone-CheckExample5">Indeterminate checkbox using custom class</label>
     </div>
+  </div>
+  <div class="d-checkbox-group">
+    <div class="d-checkbox__input">
+      <input class="d-checkbox" type="checkbox" name="Dialtone-CheckExample6" id="Dialtone-CheckExample6" indeterminate />
```
dbecherdialpad: I don't think this is valid. There is no `indeterminate` attribute, you actually can ONLY set this through JS: https://css-tricks.com/indeterminate-checkboxes/

lib/build/less/components/radio-checkbox.less
```diff
@@ -169,6 +169,18 @@
                 background-image: url("data:image/svg+xml,%3Csvg width='12' height='12' viewBox='0 0 24 24' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M9 16.2L4.8 12l-1.4 1.4L9 19 21 7l-1.4-1.4L9 16.2z' fill='hsl(240, 10%, 51%)'/%3E%3C/svg%3E");
             }
         }
+
+        &--indeterminate,
+        &[indeterminate] {
```
dbecherdialpad: as above, this is not an attribute. But there IS a CSS pseudo-class weirdly (if it's set via js) https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate

lib/build/less/components/radio-checkbox.less
```diff
@@ -169,6 +169,18 @@
...
+            --check-radio--bgc:     var(--check-radio--color);
+            background-image: url("data:image/svg+xml,%3Csvg width='12' height='12' viewBox='0 0 24 24' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M19 13H5v-2h14v2z' fill='%23fff'/%3E%3C/svg%3E");
+
+            &[disabled] {
```
dbecherdialpad, line 178: unrelated to your PR but I just noticed we aren't using the pseudo class for this either https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled

dbecherdialpad (changes requested):

docs/components/checkbox.html
dbecherdialpad: dammit @dialpad-drew already beat me to it

modialpad:
> Only question I have is about the attribute. [MDN suggests](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox) that browsers don't currently support `indeterminate` as an attribute and the example using the attribute doesn't seem to exhibit any functionality. Is there a CSS-only way to do this, or are we simply just putting in some styles for the Vue component to leverage?

@dialpad-drew @dbecherdialpad I read the docs and you're right, browsers don't support `indeterminate` attribute on input elements. I removed the attribute selector and updated this select on the `indeterminate` pseudo-class, in addition to the custom class CSS approach (`--indeterminate`). Removed the attribute selector example in the docs. 


image

lib/build/less/components/radio-checkbox.less
modialpad: Yeah, the other PR I have adds the custom class for the `disabled` state for accessibility reasons.

lib/build/less/components/radio-checkbox.less
modialpad: Done

docs/components/checkbox.html
modialpad: Done

dbecherdialpad (approved):

modialpad:
@dbecherdialpad we probably don't need to update `dialtone-vue` for this, right? Considering that the consumer won't bother updating the `indeterminate` attribute through JS. And if it's just a class, then we can do that just through dialtone. 

Also, another question I have is about the publish process. I assume I'll have to bump the version of Dialtone after this merge, is there documentation somewhere for this?

(Finally, I can't seem to have merge permissions here. But we should probably go ahead and merge if it looks good)

dbecherdialpad:
@modialpad

> we probably don't need to update dialtone-vue for this, right? Considering that the consumer won't bother updating the indeterminate attribute through JS.

For dialtone-vue I generally prefer adding a prop (which adds the class) rather than relying on engineers to pass the correct class name. This also allows for improved autocomplete and documentation. For `indeterminate` we can also provide this as a prop and have the vue component correctly set the `indeterminate` state of the checkbox (essentially implementing what you thought could be done in html).

> Also, another question I have is about the publish process. I assume I'll have to bump the version of Dialtone after this merge, is there documentation somewhere for this?

Docs are here but it requires one of us to publish the version https://github.com/dialpad/dialtone/blob/staging/.github/RELEASING.md

dbecherdialpad:
@modialpad In dialtone-vue this is actually quite trivial to add using the `.prop` v-bind modifier: https://github.com/vuejs/vue/issues/4094#issuecomment-261659740

https://vuejs.org/v2/api/#v-bind

modialpad:
> @modialpad In dialtone-vue this is actually quite trivial to add using the `.prop` v-bind modifier: [vuejs/vue#4094 (comment)](https://github.com/vuejs/vue/issues/4094#issuecomment-261659740)
> 
> https://vuejs.org/v2/api/#v-bind

Great, thanks! I'll open the `dialtone-vue` PR once Dialtone is published.

braddialpad (approved):

PR #462, opened 2021-09-29
prepublish addition. remove extra build from CI

braddialpad:
no longer need npm run build in CI as we do it as part of prepublish.

added npm run svg to prepublishOnly. npm run svg does NOT run when you execute npm run build. This was probably a big point of confusion in the manual release last night.

removed svg from setup:ci since we now run it in prepublishOnly

### DIFF
--- a/docs/getting-started/usage.html
+++ b/docs/getting-started/usage.html
@@ -27,7 +27,7 @@ description: A general overview of Dialtone's utility classes, CSS components, a
 </section>
 <section class="d-stack16">
   {% header "h2", "Components" %}
-  {% paragraph %}There are two methods to implement Dialtone components: Vue (recommended) and CSS. Vue is the preferred method as it's more robust and readily accessible out-of-the-box. <a href="https://handset.site/" class="d-link" target="_blank">Get started with Vue components.</a>{% endparagraph %}
+  {% paragraph %}There are two methods to implement Dialtone components: Vue (recommended) and CSS. Vue is the preferred method as it's more robust and readily accessible out-of-the-box. <a href="https://vue.dialpad.design/" class="d-link" target="_blank">Get started with Vue components.</a>{% endparagraph %}
   {% paragraph %}In the event Dialtone Vue doesn't suit your needs, Dialtone's CSS library offers the same set of components. These may require more work to implement and make accessible, but will work in a pinch.{% endparagraph %}
   {% codeWell %}
     {% codeWellHeader %}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -41,9 +41,6 @@ jobs:
       - name: Install dependencies
         run: npm run setup:ci
 
-      - name: Build project
-        run: npm run build
-
       - name: Deploy production
         if: ${{ github.ref == 'refs/heads/production' }}
         run: npm publish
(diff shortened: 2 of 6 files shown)


dialpad-drew:
Projects won't successfully build without the SVGs, right? If that's the case, does that suggest not running `gulp svg` at setup could mean new projects won't successfully build?

I'm thinking about this from the perspective of a product engineer who's cloned the project and set it up for the first time so they can contribute. Does this work for them?

braddialpad:
> Projects won't successfully build without the SVGs, right? If that's the case, does that suggest not running `gulp svg` at setup could mean new projects won't successfully build?
> 
> I'm thinking about this from the perspective of a product engineer who's cloned the project and set it up for the first time so they can contribute. Does this work for them?

there's a npm run setup command that is meant to be run when you initially set up the project. I still don't really like this because it goes outside the standard process most developers are used to which is "npm install" and then "npm run build" or "npm run watch". Not to mention you have to KNOW to run npm run setup again whenever there's any changes to SVGs.

Personally I think it may be best to include the svg as part of the build. Guessing this was set up as an independent script originally for performance reasons. I'll do some build time comparisons so we can make a decision.

dialpad-drew:
> > Projects won't successfully build without the SVGs, right? If that's the case, does that suggest not running `gulp svg` at setup could mean new projects won't successfully build?
> > I'm thinking about this from the perspective of a product engineer who's cloned the project and set it up for the first time so they can contribute. Does this work for them?
> 
> there's a npm run setup command that is meant to be run when you initially set up the project. I still don't really like this because it goes outside the standard process most developers are used to which is "npm install" and then "npm run build" or "npm run watch". Not to mention you have to KNOW to run npm run setup again whenever there's any changes to SVGs.
> 
> Personally I think it may be best to include the svg as part of the build. Guessing this was set up as an independent script originally for performance reasons. I'll do some build time comparisons so we can make a decision.

Dang, I just noticed you removed the SVG task from `setup:ci`, not `setup` entirely. In that case, I think my question is actually moot because the situation I described shouldn't occur if someone is following the install directions. 🤦

braddialpad:
> Dang, I just noticed you removed the SVG task from `setup:ci`, not `setup` entirely. In that case, I think my question is actually moot because the situation I described shouldn't occur if someone is following the install directions. 🤦

Yeah the same install docs should still be valid

docs/getting-started/usage.html
```diff
@@ -27,7 +27,7 @@
 </section>
 <section class="d-stack16">
   {% header "h2", "Components" %}
-  {% paragraph %}There are two methods to implement Dialtone components: Vue (recommended) and CSS. Vue is the preferred method as it's more robust and readily accessible out-of-the-box. <a href="https://handset.site/" class="d-link" target="_blank">Get started with Vue components.</a>{% endparagraph %}
+  {% paragraph %}There are two methods to implement Dialtone components: Vue (recommended) and CSS. Vue is the preferred method as it's more robust and readily accessible out-of-the-box. <a href="https://vue.dialpad.design/" class="d-link" target="_blank">Get started with Vue components.</a>{% endparagraph %}
```
braddialpad, line 30: found this sneaky boi while updating the docs

braddialpad:
> Personally I think it may be best to include the svg as part of the build. Guessing this was set up as an independent script originally for performance reasons. I'll do some build time comparisons so we can make a decision.

Okay so after testing this the build time of `npm run svg` is completely negligable (like 1 second), so I've just included it as part of the default gulp build. Makes things MUCH simpler. Updated docs/ci/netlify as well.

dbecherdialpad (approved):

docs/getting-started/usage.html
ryandialpad: Good find!

ryandialpad (approved):

PR #464, opened 2021-10-01
Improve gulp watch

braddialpad:
This PR does two things:

- fixes the issue where gulp was rebuilding infinitely on change
- greatly improves watch build performance

infinite rebuild was caused by some build generated files not being excluded from the watch, so they would change on build, be detected as a new change, and it would rebuild again

greatly improved watch rebuild performance from 27 seconds to about 3-4 seconds. Did this by a number of methods:

- Implemented gulp-cached into the gulp build. Meaning it only rebuilds files that have changed rather than all of them.
- Run the eleventy --watch --incremental task to only build changed files (same kind of deal as gulp-cached but for eleventy)
- Run gulp watch and eleventy watch in parallel for improved performance.

Note that all of these improvements are on REBUILD. the initial build will still be about the same (27 sec)

### DIFF
--- a/docs/getting-started/usage.html
+++ b/docs/getting-started/usage.html
@@ -42,7 +42,7 @@ description: A general overview of Dialtone's utility classes, CSS components, a
 </section>
 <section class="d-stack16">
   {% header "h2", "Writing CSS" %}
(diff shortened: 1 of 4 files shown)


dbecherdialpad (approved):
nice find!

dialpad-drew (approved):
Oh yeah, that's the stuff! Nice. 🏎️

PR #470, opened 2021-10-06
Updated font, border and background colors of disabled styles

juliodialpad:
## Description
The problem is we often disable inputs while data is saving data to the server, so this new disabled style is more jarring now.

https://switchcomm.atlassian.net/browse/DT-168

## Pull Request Checklist

### DIFF
--- a/docs/components/select.html
+++ b/docs/components/select.html
@@ -55,7 +55,7 @@ storybook-url: https://vue.dialpad.design/?path=/story/forms-select-menu--defaul
               <div class="d-label">
                 <label for="Dialtone--SelectExample2">Label</label>
               </div>
-              <div class="d-select">
+              <div class="d-select d-select--disabled">
                 <select class="d-select__input" id="Dialtone--SelectExample2" disabled>
                   <option value="" selected>Please select one</option>
                   <option value="option1">Option 1</option>
@@ -85,7 +85,7 @@ storybook-url: https://vue.dialpad.design/?path=/story/forms-select-menu--defaul
   <div class="d-label">
     <label for="Dialtone--SelectExample2">...</label>
   </div>
-  <div class="d-select">
+  <div class="d-select d-select--disabled>
     <select class="d-select__input" id="Dialtone--SelectExample2" disabled>
       <option value="" selected>...</option>
       <option value="...">...</option>
--- a/docs/components/select.html
+++ b/docs/components/select.html
@@ -85,7 +85,7 @@ storybook-url: https://vue.dialpad.design/?path=/story/forms-select-menu--defaul
   <div class="d-label">
     <label for="Dialtone--SelectExample2">...</label>
   </div>
-  <div class="d-select d-select--disabled>
+  <div class="d-select d-select--disabled">
     <select class="d-select__input" id="Dialtone--SelectExample2" disabled>
       <option value="" selected>...</option>
       <option value="...">...</option>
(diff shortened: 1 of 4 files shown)


dialpad-drew:
Something I didn't catch in the design: can we make these arrows on the right the same color as the text?

Screen Shot 2021-10-06 at 4 46 06 PM

juliodialpad:
In order to get that done, I'll need to add a class to dialtone **d-select--disabled**, it's ok if I do it this way? 

It's hard to style the arrow based on disabled state because the disabled state is on a children of d-select and the :before and :after that represent the arrows are siblings of .d-select__input.

image

dialpad-drew:
Hmmm, [CSS Tricks](https://css-tricks.com/almanac/selectors/d/disabled/) says you can chain these, though I haven't been able to get it to work.

> :disabled can be “chained” with other pseudo-selectors: like :before or :after to generate icons or text to provide more user feedback.

juliodialpad:
> Hmmm, [CSS Tricks](https://css-tricks.com/almanac/selectors/d/disabled/) says you can chain these, though I haven't been able to get it to work.
> 
> > :disabled can be “chained” with other pseudo-selectors: like :before or :after to generate icons or text to provide more user feedback.

The real problem here is that you cannot style a parent element based on child status, I've been investigating and seems the only way to do that is javascript, or adding a class like I did

dialpad-drew:
I'm ok w/ this. I'd like to hold off and get @braddialpad 's 👍  or 👎 before merging. Also, we'll need to update the Vue component as well with this new class, so I'll make a story for that.

dialpad-drew:
> > Hmmm, [CSS Tricks](https://css-tricks.com/almanac/selectors/d/disabled/) says you can chain these, though I haven't been able to get it to work.
> > > :disabled can be “chained” with other pseudo-selectors: like :before or :after to generate icons or text to provide more user feedback.
> 
> The real problem here is that you cannot style a parent element based on child status, I've been investigating and seems the only way to do that is javascript, or adding a class like I did

BTW, is the class you added pushed up? I don't see the style update in the deploy preview.

juliodialpad:
> > > Hmmm, [CSS Tricks](https://css-tricks.com/almanac/selectors/d/disabled/) says you can chain these, though I haven't been able to get it to work.
> > > > :disabled can be “chained” with other pseudo-selectors: like :before or :after to generate icons or text to provide more user feedback.
> > 
> > 
> > The real problem here is that you cannot style a parent element based on child status, I've been investigating and seems the only way to do that is javascript, or adding a class like I did
> 
> BTW, is the class you added pushed up? I don't see the style update in the deploy preview.

I was waiting for comments to push it, I'll push it so you can see it.

dialpad-drew (approved):
LGTM, but want to wait for Brad.

docs/components/select.html
```diff
@@ -85,7 +85,7 @@
   <div class="d-label">
     <label for="Dialtone--SelectExample2">...</label>
   </div>
-  <div class="d-select">
+  <div class="d-select d-select--disabled>
```
braddialpad: Missed a quote here.

braddialpad:
> In order to get that done, I'll need to add a class to dialtone **d-select--disabled**, it's ok if I do it this way? 
 

yep this is fine IMO

braddialpad (changes requested):

docs/components/select.html
juliodialpad: Fixed, thanks

braddialpad (approved):

PR #473, opened 2021-10-13
Dev server caching/building improvements

braddialpad:
## Description

fixed the caching to now properly rebuild less files when they are changed. Had to make a number of changes to do this.
- added import references to all the dialtone less files that reference variables from other files
- now pass in each less file source individually instead of one master file
- pass the files in in the same order that was in the master dialtone.css in case this could cause issues
- pass into gulp-progeny which crawls dependencies and rebuilds those that are needed
- after building the less, concat into a single css file so we get the same output as before the change
- transform which file is used in the head (.min or non .min) via ELEVENTY_ENV env var

all these changes are only applicable to the dev server.
additionally we no longer postcss / minify when running on dev as this is very expensive and un-necessary.
very fast rebuild, only a couple secs

## Pull Request Checklist


lib/build/less/components/banner.less
```diff
@@ -1,3 +1,5 @@
+@import (reference) "../utilities/spacing.less";
```
dbecherdialpad, line 1: this seems like it will be tricky, because it's not technically needed, but rebuilding won't work without it, is that right? I suspect people will edit these files and forget to add the import then be confused why rebuilding doesn't work

dbecherdialpad (approved):
maybe time to just use webpack

lib/build/less/components/banner.less
ryandialpad: Maybe something we can add the the development docs?

ryandialpad (approved):

lib/build/less/components/banner.less
braddialpad: >  I suspect people will edit these files and forget to add the import then be confused why rebuilding doesn't work

It would also error on the initial local dev build, just wouldn't on the prod build. We should really catch this in 99.9% of cases I think unless someone pushes without ever building locally. Also could maybe set up a linter.

braddialpad:
> maybe time to just use webpack

yeah.. was thinking about it

PR #482, opened 2021-10-22
Re-exported and added some SVGs.

dialpad-drew:
## Description
Related to https://switchcomm.atlassian.net/browse/UC-11686, which is a defect in which one of our arrow SVGs isn't properly taking its parent's fill. This PR adds re-exported SVG assets with the correct fill values.


braddialpad:
have we tested that it is now working correctly after the change?

Reason I ask is that these are the replacements I see in the gulp file

```js
.pipe(replace(' fill="none"', ''))
.pipe(replace(' fill="#000"', ''))
.pipe(replace(' fill="#141721"', ''))
```

I don't see one for fill="black" but maybe that doesn't matter?

dialpad-drew:
Yea, 

> have we tested that it is now working correctly after the change?
> 
> Reason I ask is that these are the replacements I see in the gulp file
> 
> ```js
> .pipe(replace(' fill="none"', ''))
> .pipe(replace(' fill="#000"', ''))
> .pipe(replace(' fill="#141721"', ''))
> ```
> 
> I don't see one for fill="black" but maybe that doesn't matter?

I'm not entirely sure _how_ this works, but I wonder if the empty `fill` fallsback to `black`, which[ is the initial value](https://www.w3.org/TR/SVG/painting.html#FillProperties).

FWIW, I see `black` in a few of [version 5's SVGs](https://github.com/dialpad/dialtone/blob/version5/lib/build/svg/system/add.svg?short_path=3c3acb9) and this was how Josh recommended doing it.

...and it seems to work as intended:
Screen Shot 2021-10-22 at 3 21 18 PM

braddialpad:
Alright cool. I figured it might be something like that 👍

braddialpad (approved):

PR #490, opened 2021-10-29
feat: add less/css linter to Dialtone

nikitadialpad:
## Add less/css linter to Dialtone

It's the first iteration of changes which adding `stylelint` to Dialtone.

## Pull Request Checklist

giphy -  - 2021-10-28 22-16-46


.stylelintignore
```diff
@@ -0,0 +1,2 @@
+lib/build/less/utilities
```
nikitadialpad, line 1: It's syntax error during lint utilities folder. Iteration declarations and functions lint incorrectly. Added to ignore for now

.stylelintignore
```diff
@@ -0,0 +1,2 @@
...
+lib/build/less/dialtone-reset.less
```
nikitadialpad, line 2: it's common reset file, I would propose to keep it as it is in ignore. Otherwise we will need to adjust normalize css with stylelint ignore declaration

stylelint.config.js
```diff
@@ -0,0 +1,36 @@
...
+    "at-rule-no-unknown": [
+      true,
+      {
+        "ignoreAtRules": ["/^ff-/"]
+      }
+    ],
+    "property-no-unknown": null,
+
+    "no-descending-specificity": null, // turn off for now, descending cases could be fixed on the last refactoring
+    "function-calc-no-unspaced-operator": null, // otherwise it will be an error during build
```
nikitadialpad, line 21: after run `stylelint autofix` there's a build error. I turned it off for now, I'm thinking it could be avoided in less with `stylelint ignore` declaration or be turned off like it's now

stylelint.config.js
```diff
@@ -0,0 +1,35 @@
...
+    'alpha-value-notation': 'number',
+    'max-empty-lines': 3,
+    'max-line-length': 150,
+
+    // override stylelint-config-standard rules
```
nikitadialpad, line 27: some of those rules are not applicable because of less syntax specific or output (after autofix) breaks build script, because it's invalid.
Need to play more with rules to define them specifically, like config object. But I'm not sure it's needed in our case.

.stylelintignore
braddialpad: is the syntax just not supported by stylelint?

.stylelintignore
braddialpad: yeah it's fine to keep this ignored.

stylelint.config.js
braddialpad: hmm what exactly is the error here? according to doc these should be autofixable. Is it incompatibility issues with less?
https://stylelint.io/user-guide/rules/list/function-calc-no-unspaced-operator/

stylelint.config.js
braddialpad: alright yeah I think it's okay to leave these for now. Not looking for perfection right away, we can modify the rules as we figure out what we like

lib/build/less/components/input.less
```diff
@@ -1,5 +1,5 @@
-@import (reference) '../variables/icons.less';
-@import (reference) '../variables/sizes.less';
+@import (reference) "../variables/icons.less";
```
braddialpad: can we change this rule to single quote instead of double quote? Just for consistiency with dialtone-vue and ubervoice/uberconf

stylelint.config.js
braddialpad: Seems a plugin for less in stylelint exists. Not sure if this would solve all our problems but might be worth trying:
https://www.npmjs.com/package/stylelint-less

.stylelintignore
nikitadialpad: I think so, it's a syntax error during stylelint. This error still exists if turn off all styles rules and plugins.
 
Screenshot 2021-11-09 at 15 58 20

All different options which I've tried to set up recognition for those declarations in utilities folders didn't fix it.

lib/build/less/components/input.less
nikitadialpad: done

stylelint.config.js
nikitadialpad: - `function-calc-no-unspaced-operator` could have 3 options : `true`, `false` or `null`. When it's `true` or `false`, for example, this code after lint looks like 
Screenshot 2021-11-09 at 16 10 20
because of it there's an error during `build`
Screenshot 2021-11-09 at 16 11 04

hm, I see any option to turn off this rule, what do you think about this? 

- added that plugin, the result is the same.

.stylelintignore
nikitadialpad: according to the discussion about this case, added [lesshint](https://github.com/lesshint/lesshint/blob/master/lib/linters/README.md) for utilities folder

.lesshintrc.js
```diff
@@ -0,0 +1,31 @@
+module.exports = {
+  excludedFiles: [
+    'lib/build/less/utilities/internals.less'
```
nikitadialpad, line 3: exclude this file, even lesshist is not able to read and align it

braddialpad (approved):
This is awesome, thanks!

PR #498, opened 2021-11-09
fix: bad syntax in vue files when style tags are used in SVG

braddialpad:
prior to this change if there were style tags within an SVG, when we are building the vue file they would be copied over and embedded within the template. This is erroneous syntax in vue. This change will use regex to extract style tags and then put them at the end of the vue file after the template tags have already been inserted as they should be in a vue SFC.

## Pull Request Checklist

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -31,6 +31,7 @@ var cache = require('gulp-cached');
 var order = require('gulp-order');
 var concat = require('gulp-concat');
 var remember = require('gulp-remember');
+var through2 = require('through2');
 var browsersync = require('browser-sync').create();
 var package = require('./package.json');
 
@@ -363,6 +364,8 @@ var buildSystemSVGs = function(done) {
         .pipe(dest(paths.svgs.sysOutputDocs))
         .pipe(replace('<svg', '<template>\n  <svg'))
         .pipe(replace('</svg>', '</svg>\n</template>'))
+        // move any style tags within the svg into style tags of the vue component
+        .pipe(through2.obj(moveStyleTagsToEOF))
         .pipe(rename(function(file) {
             var converted = file.basename.replace(/\b\S/g, t => t.toUpperCase()).replace(/[-]+/g, '');
 
@@ -373,6 +376,20 @@ var buildSystemSVGs = function(done) {
 
     done();
 };
+
+const moveStyleTagsToEOF = function(file, enc, cb) {
+    if (file.isBuffer()) {
+        const styleTagsRegex = /<style[\s\S]*<\/style>/gmi;
+        let code = file.contents.toString();
+        const result = styleTagsRegex.exec(code);
+        if (!result) return cb(null, file);
+        const matchedText = result[0];
+        code = code.replace(styleTagsRegex, '');
+        code = code + matchedText;
+        file.contents = Buffer.from(code)
+    }
+    return cb(null, file);
+}
 
 var buildBrandSVGs = function(done) {
 
@@ -414,6 +431,8 @@ var buildBrandSVGs = function(done) {
         .pipe(dest(paths.svgs.brandOutputDocs))
         .pipe(replace('<svg', '<template>\n  <svg'))
         .pipe(replace('</svg>', '</svg>\n</template>'))
+        // move any style tags within the svg into style tags of the vue component
+        .pipe(through2.obj(moveStyleTagsToEOF))
         .pipe(rename(function(file) {
             var converted = file.basename.replace(/\b\S/g, t => t.toUpperCase()).replace(/[-]+/g, '');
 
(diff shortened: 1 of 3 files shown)


dbecherdialpad:
imo allowing styles for SVG elements is a bit dangerous. _Usually_ these generated "class names" are pretty generic and easily clash with other SVGs, eg the play store ones are `.cls-1` etc. Why not just enforce consistency in the SVGs by using inline fills?

josedialpad:
The migration of the classes to the `style` section worked well, though have this error trying to use in `dialtone-vue`:

```
LOG from GenerateFileWebpackPlugin
<i> [GenerateFileWebpackPlugin] [version.txt] [generated]

ERROR in /Users/jose/Code/dialtone/lib/dist/vue/icons/IconPlayStoreBadge.vue?vue&type=style&index=0&lang=css& (./node_modules/css-loader/dist/cjs.js??ref--9-1!./node_modules/vue-loader/lib/loaders/stylePostLoader.js!./node_modules/postcss-loader/dist/cjs.js??ref--9-2!./node_modules/css-loader/dist/cjs.js??ref--14-1!./node_modules/vue-loader/lib??vue-loader-options!/Users/jose/Code/dialtone/lib/dist/vue/icons/IconPlayStoreBadge.vue?vue&type=style&index=0&lang=css&)
Module build failed (from ./node_modules/postcss-loader/dist/cjs.js):
SyntaxError

(1:1) /Users/jose/Code/dialtone/lib/dist/vue/icons/IconPlayStoreBadge.vue Unknown word

> 1 | // Imports
    | ^
  2 | var ___CSS_LOADER_API_IMPORT___ = require("../../../../../dialtone-vue/storybook/node_modules/css-loader/dist/runtime/api.js");
  3 | exports = ___CSS_LOADER_API_IMPORT___(true);

 @ /Users/jose/Code/dialtone/lib/dist/vue/icons/IconPlayStoreBadge.vue?vue&type=style&index=0&lang=css&
```

It seems that the CSS Loader in dialtone-vue is not working?

Another thing that noticed is that not all the classes were added when compiling, for instance in the [play-store-badge.svg](https://github.com/dialpad/dialtone/blob/staging/lib/build/svg/brand/play-store-badge.svg) icon, I see:

```
<style>.cls-1{fill:#a6a6a6;}.cls-10,.cls-2,.cls-3{fill:#fff;}.cls-2{stroke:#fff;stroke-miterlimit:10;stroke-width:0.2px;}.cls-4{fill:url(#linear-gradient);}.cls-5{fill:url(#linear-gradient-2);}.cls-6{fill:url(#linear-gradient-3);}.cls-7{fill:url(#linear-gradient-4);}.cls-8{opacity:0.2;}.cls-10,.cls-8,.cls-9{isolation:isolate;}.cls-9{opacity:0.12;}.cls-10{opacity:0.25;}</style>
```

But when building it generates:

```
<style>.cls-2{fill:#fff;stroke:#fff;stroke-miterlimit:10;stroke-width:.2px}.cls-9{isolation:isolate;opacity:.12}</style>
```

That's happening on `staging` too.

braddialpad:
> imo allowing styles for SVG elements is a bit dangerous. _Usually_ these generated "class names" are pretty generic and easily clash with other SVGs, eg the play store ones are `.cls-1` etc. Why not just enforce consistency in the SVGs by using inline fills?

This is a very good point... We should probably not allow these due to high chance of conflict.

braddialpad:
Created PR to inline the generic class names in the play store svg here: https://github.com/dialpad/dialtone/pull/499

I think we can still probably put this PR in. While we should not allow generic generated names I think there is value in allowing proper manually named classes. The class names should use BEM style naming as we do in our apps, which we will have to enforce through review.

braddialpad:
> The migration of the classes to the `style` section worked well, though have this error trying to use in `dialtone-vue`:

hmm seems like a config issue. I'll look into this in a bit

josedialpad:
@dbecherdialpad @braddialpad if the class names used in the icon components might have conflict, can't we use the [scoped styles](https://vue-loader.vuejs.org/guide/scoped-css.html) so each class is specific for each icon component?

braddialpad:
> Another thing that noticed is that not all the classes were added when compiling, for instance in the [play-store-badge.svg]

This actually looks like the minifier correctly doing its work. It inlined all classes that were only a single property, and combined ones that it could.

braddialpad:
> @dbecherdialpad @braddialpad if the class names used in the icon components might have conflict, can't we use the [scoped styles](https://vue-loader.vuejs.org/guide/scoped-css.html) so each class is specific for each icon component?

I think we used to use these in the product at one time but stopped using them across dialpad in place of BEM naming, so I didn't really consider using them. I'm not sure exactly the reason @dbecherdialpad could probably tell you.

dbecherdialpad:
@braddialpad @josedialpad yeah I was gonna mention scoped styles. The reasons we mostly stopped using them in ubervoice were varied but mostly related to how they did (or did not) work with child Vue components. In this case, I think it would actually be totally fine, and is a good use case for scoped styles.

braddialpad:
> It seems that the CSS Loader in dialtone-vue is not working?

Googled around a bunch, think I got the fix for this in https://github.com/dialpad/dialtone-vue/pull/189

Only a storybook issue, the library itself was building fine.

braddialpad:
added scoped to the style tag as requested

josedialpad (approved):

PR #504, opened 2021-11-15
fix: allow highlighting within code wells on doc site

braddialpad:
## Description

Code wells were not allowing highlighting which made copying code from them rather annoying. Removed the code that did this. Not sure why this was in there in the first place, doesn't make much sense.

### DIFF
--- a/docs/assets/less/dialtone-syntax.less
+++ b/docs/assets/less/dialtone-syntax.less
@@ -31,18 +31,6 @@ pre[class*="language-"] {
 	hyphens: none;
 }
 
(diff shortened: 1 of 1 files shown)


josedialpad:
Looks like it was a prism theme that was changed the background color in our styles https://gist.github.com/SErr0r/e94120c8a22d84977e9d3379f96ceff0#file-dabblet-css-L30-L40

josedialpad:
Unrelated change: https://github.com/dialpad/dialtone/blob/37acf1aa173489060b82f6f4dd591cf5982af685/docs/assets/less/dialtone-syntax.less#L49

it should be `inherit`

josedialpad (approved):

PR #516, opened 2021-12-04
feat(button): add muted styles

juliodialpad:
## Description

Add muted and muted outline button styles to Dialtone. The styles already existed on Dialtone so just updated the variables and added the examples.

## Pull Request Checklist

### DIFF
--- a/docs/components/button.html
+++ b/docs/components/button.html
@@ -163,6 +163,28 @@ storybook-url: https://vue.dialpad.design/?path=/story/components-button--defaul
       {% endcodeWell %}
     </div>
     <div class="d-stack8">
+      {% header "h3", "Muted" %}
+      {% paragraph %}The muted button style is used to communicate...{% endparagraph %}
+      {% codeWell %}
+      {% codeWellHeader %}
+      <div class="d-stack8">
+        <div>
+          <button class="d-btn d-btn--muted" type="button"><span class="d-btn__label">Place call</span></button>
+        </div>
+        <div>
+          <button class="d-btn d-btn--muted d-btn--outlined" type="button"><span class="d-btn__label">Place call</span></button>
+        </div>
+      </div>
+      {% endcodeWellHeader %}
+      {% codeWellFooter %}
+      {% highlight html linenos %}
+<button class="d-btn d-btn--muted" type="button"><span class="d-btn__label">...</span></button>
+<button class="d-btn d-btn--muted d-btn--outlined" type="button"><span class="d-btn__label">...</span></button>
+      {% endhighlight %}
+      {% endcodeWellFooter %}
+      {% endcodeWell %}
+    </div>
+    <div class="d-stack8">
       {% header "h3", "Link" %}
       {% paragraph %}Buttons can be styled as a <a href="/components/link.html">link</a> in situations for which you need the appearance of a link but behavior of a button. Using the {% code %}button{% endcode %} element provides a better accessibility experience.{% endparagraph %}
       {% codeWell %}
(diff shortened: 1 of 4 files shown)


braddialpad (approved):

dialpad-drew:
Idea for missing copy: 
> The muted button style is used to communicate non-primary actions for contexts in which the base style may not work (e.g. colored backgrounds, validation components, etc). This style’s use should be rare. When in doubt, use the [base button style](”The muted button style is used to communicate non-primary actions for contexts in which the base style may not work (e.g. colored backgrounds, validation components, etc). This style’s use should be rare. When in doubt, use the base button style.”).

Can we make sure to add muted button as one of the ~three~ four styles. That would also include moving that example up so its in the right order (see table of contents below).

Screen Shot 2021-12-06 at 4 56 39 PM

Also, I know it's scope creep, but it's small and we're already in the file. Can we remove the "cursor styles" section, update the "disabled" section's `.d-btn--disabled` example to show what was in "cursor styles" example, and update the "disabled" section's first paragraph's last sentence to read this:

> Using the class also requires `aria-disabled` and a wrapper to disable pointer events.

Screen Shot 2021-12-06 at 5 01 29 PM

@braddialpad is the sentence about the pointer events accurate? Or should we say pointer events and the click event both need to be disabled?

braddialpad:
> @braddialpad is the sentence about the pointer events accurate? Or should we say pointer events and the click event both need to be disabled?

The wrapper does not actually do anything with pointer events, it only displays a "not allowed" symbol on mouse over. d-btn--disabled is what sets pointer-events to none. We should probably still mention the click event as this would be handled in javascript

docs/components/button.html
```diff
@@ -123,9 +124,31 @@
         {% endcodeWellFooter %}
       {% endcodeWell %}
     </div>
+    <div class="d-stack8">
+      {% header "h3", "Muted" %}
+      {% paragraph %}The muted button style is used to communicate...{% endparagraph %}
```
braddialpad: We still need to update the copy here with what drew posted.

juliodialpad: Oops missed that one, thought I've updated it, should be fixed now.

docs/components/button.html
```diff
@@ -123,9 +124,31 @@
...
+      <button class="d-btn d-btn--muted d-btn--outlined" type="button"><span class="d-btn__label">...</span></button>
+      {% endhighlight %}
+      {% endcodeWellFooter %}
+      {% endcodeWell %}
+    </div>
     <div class="d-stack8">
       {% header "h3", "Disabled" %}
-      {% paragraph %}Buttons can be disabled using either the {% code %}disabled{% endcode %} attribute or a Dialtone class. Use the attribute when a button should appear disabled and not recieve focus; use the class when a button should appear disabled but still recieve focus (i.e. a disabled button with a tooltip). Using the class also requires {% code %}aria-disabled{% endcode %} and additional implementation to prevent the click event.{% endparagraph %}
+      {% paragraph %}Buttons can be disabled using either the {% code %}disabled{% endcode %} attribute or a Dialtone class. Use the attribute when a button should appear disabled and not recieve focus; use the class when a button should appear disabled but still recieve focus (i.e. a disabled button with a tooltip). Using the class also requires {% code %}aria-disabled{% endcode %} and a wrapper to disable pointer events.{% endparagraph %}
```
braddialpad: Let's change the last sentence to: Using the class also requires aria-disabled and a wrapper to display the "not allowed" pointer. Additional javascript implementation is required to prevent the click event.

braddialpad (changes requested):

braddialpad (approved):

PR #522, opened 2021-12-17
fix: generate missing hover/focus classes

### LINKED ISSUE #491: h:d-bgc-transparent doesn't exist for
In reading through Dialtone's utility classes, I see that we can prefix any background color with `h:` to apply hover-only styles. However I noticed that `h:d-bgc-transparent` doesn't work (if we want to negate a hover style). **Can we create this?**

There might be other styles that don't have matching hover (`h:`) and focus (`:f`) styles. I didn't do a thorough audit but wanted to make sure I documented this one case before I forgot :)

braddialpad:
Any classes that were generated via d-internals were getting h: and f: classes, however there are a bunch of classes in the system that are not generated via d-internals that were not getting any hover focus states. Created a new function `#d-internals #generate-hover-focus` where you can pass in the name of the class and the associated mixin and it will copy the properties to the h: and f: classes for these one off cases.

The following util classes should have f: and h: states based on the code and documentation:

- d-bgc
- d-bc
- d-bgg
- d-bgo
- d-bco
- d-bs
- d-fc
- d-fco

Found a bunch of stuff like `h:d-bs-lg` used in the dialtone docs which seemed to never exist? So I added h: and f: to box-shadow as well, and added to documentation page.

### DIFF
--- a/lib/build/less/utilities/colors.less
+++ b/lib/build/less/utilities/colors.less
@@ -39,69 +39,85 @@
 #d-internals #color-classes(red-700);
 
 //  --  We can't use the mixin for this one class :(
-.d-fc-primary,
-.h\:d-fc-primary:hover,
-.f\:d-fc-primary:focus {
+.d-fc-primary {
     color: hsla(var(--primary-color-h) var(--primary-color-s) var(--primary-color-l)  ~' / ' var(--fco)) !important;
 }
-
-.f\:d-fc-primary:focus-within {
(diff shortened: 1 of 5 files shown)


lib/build/less/utilities/colors.less
```diff
@@ -39,69 +39,85 @@
 #d-internals #color-classes(red-700);
 
 //  --  We can't use the mixin for this one class :(
```
josedialpad: nit: remove comment, we can use now the new mixin

josedialpad (approved):
LGTM! I can see in the build output now the utility classes that were missing with hover, focus and focus-within 👌

PR #524, opened 2021-12-29
feat: implement focus visible

juliodialpad:
## Description
We need to use `:focus-visible` instead of `:focus`: to allow keyboard navigation to still use visible focus, but remove instances in which pointer interaction shows focus indication.

## Pull Request Checklist

### DIFF
--- a/lib/build/less/components/link.less
+++ b/lib/build/less/components/link.less
@@ -50,7 +50,6 @@
       box-shadow: none;
     }
 
-
     &:focus-visible {
         border-radius: 4px;
         outline: none;
--- a/lib/build/less/components/link.less
+++ b/lib/build/less/components/link.less
@@ -40,7 +40,7 @@
         cursor: pointer;
     }
 
-    &:focus {
+    &:focus-visible {
         border-radius: 4px;
         outline: none;
         box-shadow: 0 0 0 var(--su2) var(--focus-ring);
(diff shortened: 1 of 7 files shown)


lib/build/less/utilities/internals.less
```diff
@@ -141,6 +153,9 @@
       .f\:@{className}:focus-within {
         @mixin();
       }
+      .f\:@{className}:focus-visible {
+        @mixin();
+      }
```
braddialpad: this should be .fv: right?

juliodialpad: Right, fixed it good catch, thanks!

braddialpad (approved):

nikitadialpad (approved):

lib/build/less/components/link.less
```diff
@@ -41,6 +41,17 @@
...
     &:focus {
+      border-radius: 4px;
+      outline: none;
+      box-shadow: 0 0 0 var(--su2) var(--focus-ring);
+    }
+
+    &:focus:not(:focus-visible) {
+      box-shadow: none;
+    }
+
```
nikitadialpad, line 52: there're two empty lines

nikitadialpad:
seems corresponding polyfills should be add separately, to not forget about that are we going to add some documentation for this case?

lib/build/less/components/link.less
juliodialpad: Thanks, fixed!

juliodialpad:
> seems corresponding polyfills should be add separately, to not forget about that are we going to add some documentation for this case?

Yeah, we have a ticket to add focus-visible and polyfill documentation to dialpad.design, so it's ok.

PR #528, opened 2022-01-06
feat: implement focus visible polyfill

juliodialpad:
## Description

We need to polyfill focus-visible rules to make it work with safari. 

In safari there are certain types of elements that don't trigger focus() event and so, the :focus selector never gets rendered. In order to make every browser work the same, I added a safari-only polyfill to trigger focus on elements like checkboxes and buttons [assets/js/focus-safari-polyfill.js]. This polyfill checks if the UA is safari and the attaches an event listener to capture click on several elements.

Then, the focus-visible-polyfill adds the .focus-visible class to the focused element.

Everything is working well now, but we should be careful to check that the focus-safari-polyfill doesn't consume a lot of resources as it's capturing every click event.

## Pull Request Checklist


dbecherdialpad:
Are you sure we need all this? That article is from almost 4 years ago. I believe autoprefixer is already adding the correct rules for `.focus-visible` whenever we use `:focus-visible`

juliodialpad:
> Are you sure we need all this? That article is from almost 4 years ago. I believe autoprefixer is already adding the correct rules for `.focus-visible` whenever we use `:focus-visible`

I think it might be a better approach for dialtone as we pack just css and not any js in npm package...

With this approach, we only need the polyfill in the products for the non existent components in dialtone, like dropdown menu and list item.

But if its better or the decided approach to use the polyfill in every place I can rewrite it to use it.

docs/assets/js/focus-safari-polyfill.js
```diff
@@ -0,0 +1,89 @@
+/*
```
braddialpad: So this is an additional polyfill we need on top of the focus-visible polyfill so it works in safari?

braddialpad:
Seems reasonable to me. @dbecherdialpad are you okay with this?

docs/assets/js/focus-safari-polyfill.js
juliodialpad: Yeah, that polyfill is to fix the issue that by default Safari doesn't trigger focus on certain elements like checkboxes, buttons and custom elements with tabindex set.

For example: [here](https://dialpad.design/utilities/backgrounds/color/#focus) if you click the button it wouldn't work on safari until we add the polyfill.

braddialpad: gotcha, would this have to be put in on product side as well?

dbecherdialpad: I mentioned in the other PR, I don't think it's necessarily on us to change the default focus behaviour in Safari. The focus styles _do_ work in Safari, it's just that clicking doesn't focus the element.

dbecherdialpad (commented):
this appears to change everything to use the `.focus-visible` class. Isn't the point of the polyfill that it doesn't have to run at all in browsers that support `:focus-visible` natively?

docs/assets/js/focus-safari-polyfill.js
juliodialpad: > I mentioned in the other PR, I don't think it's necessarily on us to change the default focus behaviour in Safari. The focus styles _do_ work in Safari, it's just that clicking doesn't focus the element.

Answered [here](https://github.com/dialpad/dialtone-vue/pull/222#discussion_r782413879)

juliodialpad: > gotcha, would this have to be put in on product side as well?

Yeah, we need it to standardize the behaviour across browsers

juliodialpad:
> this appears to change everything to use the `.focus-visible` class. Isn't the point of the polyfill that it doesn't have to run at all in browsers that support `:focus-visible` natively?

The focus-visible polyfill is running indistinctly on chrome or safari, so I decided to change everything to the polyfill class.

It doesn't detect if the browser supports :focus-visible or not

dbecherdialpad (approved):

braddialpad (approved):

PR #533, opened 2022-01-20
feat: add toggle styles

modialpad:
## Description

Adds the new Dialtone Toggle `d-toggle` styles. See [Figma](https://www.figma.com/file/nAvI4GmMxuzdRmZ2HdUbgs/Toggle-Proposal?node-id=309%3A33).

## Pull Request Checklist

image
DtToggle Styles Gif

### DIFF
--- a/docs/components/toggle.html
+++ b/docs/components/toggle.html
@@ -133,7 +133,7 @@ storybook-url: https://vue.dialpad.design/?path=/story/components-toggle--defaul
       {% codeWell %}
         {% codeWellHeader %}
 <fieldset class="d-stack8 d-w50p">
-  <legend class="d-label d-fs20">User Settings</legend>
+  <legend class="d-label d-fs20">DND Settings</legend>
   <div class="d-toggle-wrapper d-d-flex d-fd-column">
     <div class="row d-d-flex d-ai-center d-jc-space-between d-mb12">
       <h3>Status</h3>
@@ -185,7 +185,7 @@ storybook-url: https://vue.dialpad.design/?path=/story/components-toggle--defaul
 {% codeWellFooter %}
   {% highlight html linenos %}
 <fieldset class="d-stack8 d-w50p">
-  <legend class="d-label d-fs20">User Settings</legend>
+  <legend class="d-label d-fs20">DND Settings</legend>
   <div class="d-toggle-wrapper d-d-flex d-fd-column">
     <div class="row d-d-flex d-ai-center d-jc-space-between d-mb12">
       <h3>Status</h3>
--- a/docs/components/toggle.html
+++ b/docs/components/toggle.html
@@ -136,40 +136,40 @@ storybook-url: https://vue.dialpad.design/?path=/story/components-toggle--defaul
   <legend class="d-label">User Settings</legend>
   <div class="d-toggle-group d-d-flex d-ai-center">
     <div class="d-toggle-copy d-mr6">
-      <label class="d-toggle-label" for="Dialtone-Toggle1">Unchecked Toggle</label>
+      <label class="d-toggle-label" for="Dialtone-Toggle5">Unchecked Toggle</label>
     </div>
     <div class="d-toggle-button">
-      <button class="d-toggle" type="button" role="switch" id="Dialtone-Toggle1" >
+      <button class="d-toggle" type="button" role="switch" id="Dialtone-Toggle5" >
         <span class="d-toggle__inner" />
       </button>
     </div>
   </div>
   <div class="d-toggle-group d-d-flex d-ai-center">
     <div class="d-toggle-copy d-mr6">
-      <label class="d-toggle-label" for="Dialtone-Toggle2">Checked Toggle</label>
+      <label class="d-toggle-label" for="Dialtone-Toggle6">Checked Toggle</label>
     </div>
     <div class="d-toggle-button">
-      <button class="d-toggle d-toggle--checked" type="button" role="switch" id="Dialtone-Toggle2" >
+      <button class="d-toggle d-toggle--checked" type="button" role="switch" id="Dialtone-Toggle6" >
         <span class="d-toggle__inner" />
       </button>
     </div>
   </div>
   <div class="d-toggle-group d-d-flex d-ai-center">
     <div class="d-toggle-copy d-mr6">
-      <label class="d-toggle-label" for="Dialtone-Toggle3">Unchecked Disabled</label>
+      <label class="d-toggle-label" for="Dialtone-Toggle7">Unchecked Disabled</label>
     </div>
     <div class="d-toggle-button">
-      <button class="d-toggle" type="button" role="switch" disabled id="Dialtone-Toggle3" >
+      <button class="d-toggle" type="button" role="switch" disabled id="Dialtone-Toggle7" >
         <span class="d-toggle__inner" />
       </button>
     </div>
   </div>
   <div class="d-toggle-group d-d-flex d-ai-center">
     <div class="d-toggle-copy d-mr6">
-      <label class="d-toggle-label" for="Dialtone-Toggle4">Checked Disabled</label>
+      <label class="d-toggle-label" for="Dialtone-Toggle8">Checked Disabled</label>
     </div>
     <div class="d-toggle-button">
-      <button class="d-toggle d-toggle--checked" type="button" role="switch" disabled id="Dialtone-Toggle4" >
+      <button class="d-toggle d-toggle--checked" type="button" role="switch" disabled id="Dialtone-Toggle8" >
         <span class="d-toggle__inner" />
       </button>
     </div>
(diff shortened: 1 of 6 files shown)


docs/components/toggle.html
```diff
@@ -0,0 +1,240 @@
...
+    </div>
+  </div>
+</section>
+<section class="d-stack16">
+    {% header "h2", "Accessibility" %}
+    {% paragraph %}The best accessibility is semantic HTML. Most screen readers understand how to parse buttons if they’re correctly formatted. When it comes to toggles, there are a few things to keep in mind:{% endparagraph %}
+    {% ul %}
+      <li>All toggle buttons should have an {% code %}id{% endcode %} attribute.</li>
+      <li>Associate toggle labels with their buttons using the {% code %}for{% endcode %} attribute. This correlates with the toggle's {% code %}id{% endcode %}.</li>
+      <li>If you have a group of related toggles, use a {% code %}fieldset{% endcode %} to group them and a {% code %}legend{% endcode %} to title the group. For further information, please visit Gov.UK’s article, <a href="https://accessibility.blog.gov.uk/2016/07/22/using-the-fieldset-and-legend-elements/">"Using the fieldset and legend elements"</a>.</li>
```
tedgoas-dialpad: Does this link need a class on it? I'd expect it to be purple, but it's showing as blue. Is this an issue with the docs site? cc @braddialpad 

Screen Shot 2022-01-20 at 3 30 58 PM

docs/components/toggle.html
```diff
@@ -0,0 +1,240 @@
...
+      </button>
+    </div>
+  </div>
+</fieldset>
+          {% endhighlight %}
+        {% endcodeWellFooter %}
+      {% endcodeWell %}
+    </div>
+    <div class="d-stack8">
+      {% header "h3", "Stacked Group" %}
```
tedgoas-dialpad, line 132: This is an interesting variant on this. I wonder if we could expand this stacked pattern to left-align the labels and right-align the toggle , like so:

image


(instead of having the label and toggle squished together)

tedgoas-dialpad (commented):
One question about the stacked pattern. Component itself looks good!

docs/components/toggle.html
braddialpad: hmmm wondering if we might just want this to be the default for everything. @modialpad should be a simple change, would just need to add `d-jc-space-between` to the divs with class `d-toggle-group`

docs/components/toggle.html
braddialpad: yep it just needs `class="d-link"` added to it

docs/components/toggle.html
tedgoas-dialpad: Ooh that's a clever way to handle it, good suggestion Brad!

lib/build/less/components/toggle.less
```diff
@@ -0,0 +1,96 @@
...
+//  TABLE OF CONTENTS
+//  • BASE STYLE
+//  • VISUAL STYLES
+//  • ALTERNATE STYLES
+//
+//
+//  ============================================================================
+//  $   BASE STYLE
+//  ----------------------------------------------------------------------------
+.d-toggle {
```
braddialpad, line 20: We generally try to avoid using px values as they are not scalable. Try to use the built in dialtone stops for the specific property if available, if not you can possibly use our [spacing units](https://github.com/dialpad/dialtone/blob/staging/lib/build/less/variables/spacing.less). I will show you examples of some fixes below and let you fix the rest. 

If you look into some of the other less components in the system you'll probably see where I'm coming from.

lib/build/less/components/toggle.less
```diff
@@ -0,0 +1,96 @@
...
+  --toggle-ts: var(--td200);
+  --toggle-bc: var(--black-200);
+
+  position: relative;
+  display: inline-block;
+  box-sizing: border-box;
+  width: 48px;
+  height: 24px;
+  padding: 0;
+  line-height: 20px;
```
braddialpad: `line-height: var(--lh4);`

lib/build/less/components/toggle.less
```diff
@@ -0,0 +1,96 @@
...
+  //  Component specific CSS Vars
+  //  ------------------------------------------------------------------------
+  --toggle--ta: var(--ttf-in-out);
+  --toggle-ts: var(--td200);
+  --toggle-bc: var(--black-200);
+
+  position: relative;
+  display: inline-block;
+  box-sizing: border-box;
+  width: 48px;
```
braddialpad: `width: var(--su48)`

lib/build/less/components/toggle.less
```diff
@@ -0,0 +1,96 @@
...
+  height: 24px;
+  padding: 0;
+  line-height: 20px;
+  vertical-align: middle;
+  background-color: var(--toggle-bc);
+  border: 1px solid var(--toggle-bc);
+  border-radius: 25px;
```
braddialpad: the dialtone stop is `var(--br24)`. 1 pixel difference, guessing it's close enough to work but try it out.

lib/build/less/components/toggle.less
```diff
@@ -0,0 +1,96 @@
...
+    color: var(--white);
+    background-image: url("data:image/svg+xml;charset=UTF-8, %3csvg viewBox='0 0 24 24' xmlns='http://www.w3.org/2000/svg' %3e%3cpath d='M19 6.41L17.59 5 12 10.59 6.41 5 5 6.41 10.59 12 5 17.59 6.41 19 12 13.41 17.59 19 19 17.59 13.41 12z' fill='%23fff' /%3e%3c/svg%3e");
+    background-repeat: no-repeat;
+  }
+
+  &::after {
+    position: absolute;
+    top: 1px;
+    left: 1px;
+    width: 20px;
```
braddialpad: 20px is not a dialtone stop, but you can do `calc(var(--su16) + var(--su4))`

docs/components/toggle.html
modialpad: I updated the styles to look something like this:

image

modialpad: @tedgoas-dialpad @braddialpad Would this work?

docs/components/toggle.html
modialpad: Done

lib/build/less/components/toggle.less
```diff
@@ -0,0 +1,96 @@
...
+    height: 20px;
+    background-color: var(--white);
+    border-radius: 50%;
```
braddialpad: `border-radius: var(--br-circle)`

docs/components/toggle.html
braddialpad: probably don't need quotes around "Using the fieldset and legend elements" since it's a link

lib/build/less/components/toggle.less
```diff
@@ -0,0 +1,96 @@
...
+  &::after {
+    left: 25px;
+  }
+}
+
+//  $$  DISABLED TOGGLE
+//  ----------------------------------------------------------------------------
+.d-toggle--disabled,
+ .d-toggle[disabled] {
+  cursor: no-drop;
```
braddialpad: Is there a reason we use `no-drop` instead of `not-allowed`?

docs/components/toggle.html
braddialpad: add `class="d-link"` as Ted mentioned.

braddialpad (changes requested):
Looks pretty good! My comments are mostly related to getting rid of fixed px values.

docs/components/toggle.html
braddialpad: I think that's fine for the example on this doc site. Keep in mind for the vue component we may not want to have this on by default so the user can pass in whatever justify content class they need.

lib/build/less/components/toggle.less
modialpad: I see, I updated the file with our `--su` vars

lib/build/less/components/toggle.less
modialpad: Done

lib/build/less/components/toggle.less
modialpad: Done

lib/build/less/components/toggle.less
modialpad: I updated this to a --su48 as it was closer to what the Figma designs expected

lib/build/less/components/toggle.less
modialpad: Done

lib/build/less/components/toggle.less
modialpad: Done

docs/components/toggle.html
modialpad: Added

lib/build/less/components/toggle.less
modialpad: Seems like `not-allowed` has the same cursor, and is better semantically since `no-drop` refers to draggable elements

lib/build/less/components/toggle.less
modialpad: I made a fair bit of `calc` usage inside that file. I know it's widely supported by most browsers by now, but do we see an issue using `calc` in older or incompatible environments?

braddialpad: It has very good modern browser support, not something we need to worry about:
https://caniuse.com/calc

braddialpad (approved):
awesome thank you! I will do a release once you merge this

PR #540, opened 2022-01-31
feat: popover css and documentation

braddialpad:
## Description
CSS moved from dialtone-vue to dialtone. Many edits made for simplicity.

Documentation and examples written for popover.

Found that we probably will need non-modal popovers for some cases, however based on how we use popovers modal should be the default. 

Added a new z-index `zi-modal-element`. The `zi-modal` z-index is where the modal overlay itself renders, so if you need to render something on top of the modal overlay, as we do in this case, we use `zi-modal-element`. `zi-notification` was used previously but this is not correct as notifications should display above a modal.

Will not actually merge this to staging, going to create a beta branch on dt and dt-vue for all these breaking popover changes and that will make it easier to migrate into product.

## Pull Request Checklist


braddialpad:
Adding Jose now that he's back. @josedialpad could you review this for me? It's been sitting here for 11 days

docs/assets/js/popover.js
```diff
@@ -0,0 +1,56 @@
...
+
+  $.fn.openModal = function(e) {
+    const closestOverlay = $(e.target).closest('.d-popover').siblings('.d-modal').first();
+    const contentContainer = $(e.target).siblings('.d-popover__dialog').first();
+    closestOverlay.addClass('d-modal--animate-in');
+    closestOverlay.removeClass('d-modal--animate-out');
+    contentContainer.removeClass('d-vi-hidden');
+    contentContainer.addClass('d-vi-visible');
+    closestOverlay.attr('aria-hidden','false');
+    console.log(closestOverlay)
```
josedialpad: nit: remove console statement

lib/build/less/variables/layout.less
```diff
@@ -128,6 +129,7 @@
     zi-tooltip:                 @zi-tooltip;
     zi-drawer:                  @zi-drawer;
     zi-modal:                   @zi-modal;
+    zi-modal-element:           @zi-modal-element;
```
josedialpad: should we add this in Utilities > Layout > Z-index classes table?

lib/build/less/components/popover.less
```diff
@@ -0,0 +1,50 @@
+.d-popover__dialog {
```
josedialpad, line 22: In other components' styles, we have TOC at the beginning and sections well separated with comments. Just wondering if we want to maintain the same structure here

docs/assets/js/popover.js
```diff
@@ -0,0 +1,56 @@
+$(document).ready(function() {
+
+  const popoverAnchor = $('.js-popover-anchor');
+  const overlay = $('.d-modal');
+  const body = document.body;
+  let shown = false;
+
+  $.fn.closeModal = function() {
+    overlay.addClass('d-modal--animate-out');
```
josedialpad: don't we need to add the transition classes to popover too? Just to match with the dialtone-vue implementation. Though, in dialtone-vue the popover is not using the modal transition classes but the fade transition by default.

docs/assets/js/popover.js
```diff
@@ -0,0 +1,56 @@
...
+    overlay.removeClass('d-modal--animate-in');
+    $('.d-popover__dialog').removeClass('d-vi-visible');
+    $('.d-popover__dialog').addClass('d-vi-hidden');
+    overlay.attr('aria-hidden','true');
+    $(body).removeClass('d-of-hidden');
+    shown = !shown;
+  }
+
+  $.fn.openModal = function(e) {
```
josedialpad: nit: In the non-modal example, it seems you can close it by clicking in other popover button.

josedialpad:
Looks good! Let me ask, is this PR related with the changes made in the dialtone-vue implementation https://github.com/dialpad/dialtone-vue/pull/235? If so, needs both to be sync and released at the same time since they are breaking changes?

braddialpad:
> Looks good! Let me ask, is this PR related with the changes made in the dialtone-vue implementation [dialpad/dialtone-vue#235](https://github.com/dialpad/dialtone-vue/pull/235)? If so, needs both to be sync and released at the same time since they are breaking changes?

That's correct. I'm not going to actually merge this into staging, it will go into a feature branch

docs/assets/js/popover.js
braddialpad: Since the popover transition classes are implemented in vue, not with css classes, I'm not going to include them here.

docs/assets/js/popover.js
braddialpad: Not too concerned about this for the purposes of this example.

docs/assets/js/popover.js
braddialpad: thanks, done

lib/build/less/variables/layout.less
braddialpad: missed that thanks

lib/build/less/components/popover.less
braddialpad: Added this commenting TOC style

josedialpad (approved):

PR #560, opened 2022-03-21
docs: dialtone - github pages deploy

braddialpad:
Workflows are mostly the same as the other PR in dialtone-vue. Main change that was needed here was because github pages deploys to a subdir `https://dialpad.github.io/dialtone/` all links with `/` in them were referencing `https://dialpad.github.io/` instead of `https://dialpad.github.io/dialtone/`.

Used the [html base tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) to handle this. However unfortunately every link in the system had to be changed to remove the preceding `/` for this to work correctly. A bit annoying but it does work. We may need to change this again once we map the domain `dialpad.design` to this github pages site.


josedialpad:
Looks great and super fast. It seems we weren't caching the document in netlify and now we do in GP?

CleanShot 2022-03-21 at 18 29 33@2x

CleanShot 2022-03-21 at 18 30 17@2x

josedialpad (approved):

braddialpad:
> Looks great and super fast. It seems we weren't caching the document in netlify and now we do in GP?

Yep definitely no caching in the netlify process currently.

PR #561, opened 2022-03-22
feat: dt-407 support icons above and below the labels on buttons

felipe-dialpad:
## Description
Currently, button icons are only support to go to the left or right of the label. However, for the redesigned call bar (DP-46328), we need to support icons above the labels as well. So, I’m adding the ability to position the icons on top/bottom from the labels as well.

This will include two PRs, one in dialtone and one in dialtone-vue.

## Pull Request Checklist

corgi-dogs

See the screenshot below for the results:
Screen Shot 2022-03-22 at 2 20 00 AM

### DIFF
--- a/lib/build/less/components/button.less
+++ b/lib/build/less/components/button.less
@@ -151,9 +151,16 @@
     line-height: 0;
 }
 
-:not(.d-btn--vertical) > .d-btn__icon {
+.d-btn__icon--left,
+.d-btn__icon--right {
     margin-top: var(--sun4);
     margin-bottom: var(--sun4);
+}
+
+.d-btn__icon--top,
+.d-btn__icon--bottom {
+    margin-left: var(--sun4);
+    margin-right: var(--sun4);
 }
 
 .d-btn__icon--left,
@@ -182,6 +189,16 @@
 .d-btn__icon--right ~ .d-btn__label,
 .d-btn__label ~ .d-btn__icon--left {
     margin-right: var(--su4);
+}
+
+.d-btn__icon--top ~ .d-btn__label,
+.d-btn__label ~ .d-btn__icon--bottom {
+    margin-top: var(--su4);
+}
+
+.d-btn__icon--bottom ~ .d-btn__label,
+.d-btn__label ~ .d-btn__icon--top {
+    margin-bottom: var(--su4);
 }
 
 //  ============================================================================
--- a/lib/build/less/components/button.less
+++ b/lib/build/less/components/button.less
@@ -140,20 +140,29 @@
 }
 
 //  ============================================================================
-//  $   BUTTON ICON
+//  $   BUTTON WITH ICON
 //  ----------------------------------------------------------------------------
+.d-btn--vertical {
+    flex-direction: column;
+}
+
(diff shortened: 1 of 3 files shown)


lib/build/less/components/button.less
```diff
@@ -140,20 +140,29 @@
...
+.d-btn__icon--left,
+.d-btn__icon--top {
     order: -1;
 }
 
-.d-btn__icon--right {
+.d-btn__icon--right,
+.d-btn__icon--bottom {
     order: 1;
 }
```
braddialpad, line 174: We have `var(--su4);` of margin between the icon and the label on the left and right side, we should have the same when it's on top/bottom for consistency. Will probably have to separate out the selectors to do so, but that's okay.

lib/build/less/components/button.less
```diff
@@ -140,20 +140,29 @@
...
 
 //  ============================================================================
-//  $   BUTTON ICON
+//  $   BUTTON WITH ICON
 //  ----------------------------------------------------------------------------
+.d-btn--vertical {
+    flex-direction: column;
+}
```
braddialpad, line 147: Can probably just include this directly in `d-btn__icon--top` and `d-btn__icon--bottom` to avoid creating a new class

lib/build/less/components/button.less
```diff
@@ -140,20 +140,29 @@
...
+
 .d-btn__icon {
     display: flex;
+    line-height: 0;
+}
+
+:not(.d-btn--vertical) > .d-btn__icon {
```
braddialpad: Once we get rid of `d-btn--vertical` these margins can just be assigned to left/right directly

braddialpad (changes requested):
Thanks for this Felipe!

lib/build/less/components/button.less
felipe-dialpad: Hey @braddialpad, I can't get rid of this class because the flex-direction needs to go into the parent (the button), and not the icon itself. The button is already a flex container, with two childs (the icon and the label), and so it needs that to not be horizontally rendered.

I'll do the other changes as suggested.

lib/build/less/components/button.less
```diff
@@ -140,20 +140,36 @@
...
     margin-bottom: var(--sun4);
-    line-height: 0;
 }
 
-.d-btn__icon--left {
+.d-btn__icon--top,
+.d-btn__icon--bottom {
+    margin-left: var(--sun4);
+    margin-right: var(--sun4);
+}
```
felipe-dialpad: fwiw, I added these negative margins to match the left / right above, but I don't see how they make any difference in practice.

lib/build/less/components/button.less
braddialpad: ah yeah I see. No problem leave as is.

lib/build/less/components/button.less
braddialpad: hmm yeah not sure if this is needed but we can keep it for now.

braddialpad (approved):

braddialpad:
I'll do a release once you merge this in.

PR #563, opened 2022-04-04
feat(input): support different icon sizes

braddialpad:
## Description
https://dialpad.atlassian.net/browse/DT-359

Prior to this change setting an input icon size to anything other than the default would cause strange behaviour with alignment and centering. Updated to properly center using flexbox rather than fixed padding values. We still have to set the icon height to the height of the input for a couple reasons:

1. There's no such thing as a previous sibling selector, so we have no way of dynamically matching the icon height to the input.
2. When it's a textarea we must set the height to the height of one row, rather than the height of the entire input.

The input size class will now need to be put on the icon element in addition to the input itself (if anything other than default)

Added some new documentation examples to the input page to show how to use different sized icons.

This will require a small update to dialtone-vue as well.

## Pull Request Checklist

### DIFF
--- a/docs/components/input.html
+++ b/docs/components/input.html
@@ -422,7 +422,7 @@ storybook-url: https://vue.dialpad.design/?path=/story/components-input--default
                 </div>
                 <div class="d-input__wrapper">
                   <span class="d-input-icon d-input-icon--left d-input--sm d-input-icon--lg">{% iconSystem "send" %}</span>
-                  <input class="d-input d-input-icon--left d-input--sm" id="Dialtone--InputExample--IconLeft" type="text" placeholder="Placeholder" />
+                  <input class="d-input d-input-icon--left d-input--sm" id="Dialtone--InputExample--IconLeft-sm-lg" type="text" placeholder="Placeholder" />
                 </div>
               </div>
               <div class="d-w100p">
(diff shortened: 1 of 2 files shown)


docs/components/input.html
```diff
@@ -409,6 +409,77 @@
...
+        {% codeWell %}
+          {% codeWellHeader %}
+            <div class="d-stack24 d-w100p">
+              <div class="d-w100p">
+                <div>
+                  <label class="d-label" for="Dialtone--InputExample--IconLeft">Input:sm Icon:lg</label>
+                </div>
+                <div class="d-input__wrapper">
+                  <span class="d-input-icon d-input-icon--left d-input--sm d-input-icon--lg">{% iconSystem "send" %}</span>
+                  <input class="d-input d-input-icon--left d-input--sm" id="Dialtone--InputExample--IconLeft" type="text" placeholder="Placeholder" />
```
josedialpad: nit: use unique ids (e.g. `Dialtone--InputExample--IconLeftSm`) for these examples and match with the label.

josedialpad (approved):
w/ one comment

docs/components/input.html
braddialpad: ah yeah missed that thanks

PR #572, opened 2022-04-20
docs: upgrade docsearch to version 3. prevent subdir overlap

braddialpad:
Updated docsearch to v3 so we could use the facetFilters feature so dialpad.design does not give search results from dialpad.design/version5.

Upgrading from v2 to v3 changed the search from an input to a button. This caused a bunch of styling issues which I had to manually manipulate the dom to fix, giving it dialtone styling. Was a bit hacky but it worked.

Added facetFilters based on branch name so the page you are on only searches within that branch instead of all subdir pages on dialpad.design. Seems to work correctly, but I also know algolia only crawls data once per week or so, so not completely sure.

### DIFF
--- a/docs/_data/site.js
+++ b/docs/_data/site.js
@@ -4,7 +4,6 @@ module.exports = {
   localurl: "http://localhost:4000/",
   url: "https://dialpad.design",
   baseurl: process.env.ELEVENTY_BASE_URL !== '/' ? `/${process.env.ELEVENTY_BASE_URL}/` : '/',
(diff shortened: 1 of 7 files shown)


juliodialpad (approved):
Looks good, it's necessary to change the styling to be an input even if we can't type in it anymore?

braddialpad:
> Looks good, it's necessary to change the styling to be an input even if we can't type in it anymore?

I don't think so. It looks reasonable enough as an outline button

PR #584, opened 2022-05-09
fix: dt-436 fc-warning not mapped correctly

braddialpad:
Utility class for `fc-warning` was mapping straight to `fc-yellow-600` when it should be `hsla(var(--yellow-500-h) var(--yellow-500-s) calc(var(--yellow-500-l) - 20%) ~' / ' var(--alpha, 100%));` based on the token. Looks like this was changed in the variable but was not updated to the utility class.

### DIFF
--- a/docs/_data/type.json
+++ b/docs/_data/type.json
@@ -218,7 +218,7 @@
     },
     {
       "var": "warning",
-      "output": "yellow-600"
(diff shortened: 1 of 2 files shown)


juliodialpad (approved):
LGTM, I'll include this change in my next vuepress PR

PR #587, opened 2022-05-10
feat: add style for chip WIP

DestaDialpad:
## Description
This is still WIP. Here for some early feedback from designer and code reviewer
Figma: https://www.figma.com/file/QP2VRL6aOPGYXUAUHhqiuL/Chip-Proposal?node-id=714%3A21556

## Pull Request Checklist

### DIFF
--- a/docs/components/chip.html
+++ b/docs/components/chip.html
@@ -1,0 +1,142 @@
+---
+layout: page
+title: Chip
+description: A chip is a compact UI element that provides brief, descriptive information about an element. It is terse, ideally one word.
+---
+<section class="d-stack16">
+  {% header "h2", "Classes" %}
+  <table class="d-table">
+    <thead>
+      <tr>
+        <th scope="col" class="d-w40p">Class</th>
+        <th scope="col" class="d-w30p">Applies to</th>
+        <th scope="col">Description</th>
+      </tr>
+    </thead>
+    <tbody>
+      {% for i in chip.classes %}
+        {% assign class = i.class %}
+        {% assign apply = i.applies %}
+        {% assign desc = i.description %}
+        <tr>
+          <th scope="row" class="d-ff-mono d-fc-purple d-fw-normal d-fs12">.{{ class }}</th>
+          <td class="d-ff-mono d-fc-orange d-fs12">{% if apply != "N/A" %}<span class="code-example--inline">.{% endif %}{{ apply }}{% if apply != "N/A" %}</span>{% endif %}</td>
+          <td class="d-fs12">{{ desc }}</td>
+        </tr>
+      {% endfor %}
+    </tbody>
+  </table>
+</section>
+<section class="d-stack16">
+  {% header "h2", "Examples" %}
+  <div class="d-stack48">
+    <div class="d-stack8">
+      {% header "h3", "Base" %}
+      {% paragraph %}The base chip should be the go-to chip for most of your needs.{% endparagraph %}
+      {% codeWell %}
+        {% codeWellHeader %}
+          <span class="d-chip">
+            Chip
+            <button class="d-btn d-btn--circle" type="button">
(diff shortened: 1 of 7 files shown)


lib/build/less/components/chip.less
```diff
@@ -0,0 +1,104 @@
...
+  justify-content: center;
+  box-sizing: border-box;
+  padding: var(--su4) var(--su6);
+  color: var(--chip--fc);
+  font-size: var(--fs16);
+  font-family: inherit;
+  line-height: var(--lh4);
+  background-color: var(--chip--bgc);
+  border-radius: var(--chip--br);
+  cursor: pointer;
```
braddialpad: we'll need to make a new class `d-chip--interactive` which would enable `cursor: pointer` and hover styles

docs/components/chip.html
```diff
@@ -0,0 +1,142 @@
...
+</section>
+<section class="d-stack16">
+  {% header "h2", "Examples" %}
+  <div class="d-stack48">
+    <div class="d-stack8">
+      {% header "h3", "Base" %}
+      {% paragraph %}The base chip should be the go-to chip for most of your needs.{% endparagraph %}
+      {% codeWell %}
+        {% codeWellHeader %}
+          <span class="d-chip">
```
braddialpad: Because this is a span, and is interactive we'll need to set tabindex="0" on it. Thought about whether we could actually make this a `<button>` element when interactive, but a button within a button is invalid HTML. 

Looked into some other design systems that have this component and they also used span so I think we're okay here.

docs/components/chip.html
```diff
@@ -0,0 +1,142 @@
...
+            Chip
+            <button class="d-btn d-btn--circle" type="button">
```
braddialpad: This close button needs to announce "close" when focused via screenreader. aria-label="close" should do it.

lib/build/less/components/chip.less
```diff
@@ -0,0 +1,104 @@
...
+//  visit https://dialpad.design/components/chip
+//
+//  TABLE OF CONTENTS
+//  • BASE STYLE
+//    • ALTERNATE STYLE
+//
+//  ============================================================================
+//  $   BASE STYLE
+//  ----------------------------------------------------------------------------
+.d-chip {
```
braddialpad, line 17: In the design there is truncation at a certain length of content, doesn't seem to be applied here, maybe still WIP?

lib/build/less/components/chip.less
braddialpad: Once we have this we'll have to add an example for "non-interactive"

braddialpad (commented):
Looking really good! nice use of existing dialtone variables / naming standards. Didn't really compare styles to the design much since I know many things are still WIP. Comments are mostly about accessibility and the interactivity issues we already spoke about.

docs/components/chip.html
braddialpad: Clicking close puts the entire chip in an active state, it should only put the close button in an active state.

lib/build/less/components/chip.less
DestaDialpad: @braddialpad  Can I add it to the component?
```
    <span
      v-if="$slots.default"
      data-qa="dt-chip-label"
      class="d-truncate"
    >
      <!-- @slot Content within chip -->
      <slot />
    </span>

DestaDialpad:
Ping @juliodialpad for VuePress migration process

juliodialpad:
> Ping @juliodialpad for VuePress migration process

Thanks @DestaDialpad I'll be doing the migration of your work

juliodialpad:
We might want to close this PR as soon as we are sure the migration is correct.

@DestaDialpad @braddialpad

DestaDialpad:
Closing this as it's continued in  https://github.com/dialpad/dialtone/pull/597

PR #594, opened 2022-05-18
add active close logic

braddialpad:
It ain't pretty but it'll do the job.

### DIFF
--- a/docs/assets/js/chip.js
+++ b/docs/assets/js/chip.js
@@ -1,9 +1,16 @@
 $(document).ready(function() {
-    var closeButton = $('.js-chip-button');
+    var chip = $('.js-chip');
+    var chipBtn = $('.js-chip-button');
 
-    $(closeButton).on('click', function(e) {
-        console.log('clicked');
-        e.stopPropagation();
-        e.preventDefault();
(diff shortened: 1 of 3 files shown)


lib/build/less/components/chip.less
```diff
@@ -61,6 +49,10 @@
   line-height: 0;
 }
 
+.d-chip--active {
```
braddialpad, line 52: since regular :active pseudo class logic doesn't work here, apply this using a regular class.

docs/assets/js/chip.js
```diff
@@ -1,9 +1,15 @@
...
+    var chip = $('.js-chip');
+    var chipBtn = $('.js-chip-button');
 
-    $(closeButton).on('click', function(e) {
-        console.log('clicked');
-        e.stopPropagation();
-        e.preventDefault();
+    $(chip).on('mousedown', function(e) {
+        const target = $(e.target);
+        if (!target.parents().is(chipBtn)) {
```
braddialpad, line 7: check if the inner button was clicked, if not then apply the active class.

DestaDialpad (approved):
Thanks for putting this up. LGTM

PR #597, opened 2022-05-18
feat(chip): add chip component

juliodialpad:
## Description

@DestaDialpad created a PR https://github.com/dialpad/dialtone/pull/587 with all the changes to add chip component to dialtone... the work was started while we were making the migration to a new Docsite.

This resulted in the necessity to migrate this features to the new Docsite, this PR contains all the work from Desta, the necessary migrations and minor improvements to examples.

If there's anything missing or wrong, just let me know and I'll fix it ASAP.

## Pull Request Checklist


braddialpad:
Should we be merging into DP-420 instead of staging?

juliodialpad:
> Should we be merging into DP-420 instead of staging?

No, I cherry-picked every commit from DP-420, should be fine to merge with staging, I created this branch from staging also.

docs/.vuepress/exampleComponents/ExampleChip.vue
```diff
@@ -0,0 +1,53 @@
...
+      class="d-chip"
+      :class="{
+        'd-chip--interactive': interactive,
+        'd-w102': truncate,
+        'd-chip--active': isActive,
+        [`d-chip--${size}`]: size
+      }"
+      tabindex="0"
+      @mousedown="setActive"
+      @mouseup="setActive"
```
braddialpad, line 12: Currently if you mouse down on the chip, then move your mouse outside of the chip and release, the active state stays. I think it will fix this if you add the @mouseleave event

braddialpad:
@natayi-dialpad could you visually verify since we've done the migration. Just go here: https://dialpad.design/deploy-previews/pr-597

docs/.vuepress/exampleComponents/ExampleChip.vue
braddialpad: another edge case situation.. if you click the chip and then drag and release over the X the button stays active

natayi-dialpad:
- Close button hover colour needs to be updated.
- The icon looks a bit big (it wasn't as notice-able when in Figma prob cuz I was using the email one), let's size it down so it's the same size as the close icon.
- I don't think the target / hover area that extends out of the close button's actual size is applied yet.
- The non-interactive chip showed a focus state (that looks like the browser's default focus?)

cc: @francisrupert

docs/.vuepress/exampleComponents/ExampleChip.vue
```diff
@@ -0,0 +1,53 @@
+<template>
+  <span
+      class="d-chip"
+      :class="{
+        'd-chip--interactive': interactive,
+        'd-w102': truncate,
+        'd-chip--active': isActive,
+        [`d-chip--${size}`]: size
+      }"
+      tabindex="0"
```
braddialpad: must not be set when interactive is false

juliodialpad: I'm validating that on "setActive" function, if `!props.interactive` it doesn't set the `isActive` class to true.

docs/.vuepress/exampleComponents/ExampleChip.vue
juliodialpad: > Currently if you mouse down on the chip, then move your mouse outside of the chip and release, the active state stays. I think it will fix this if you add the @mouseleave event

That fixed it, thanks!

juliodialpad: Fixed it too, I had a validation not working as expected.

juliodialpad: > another edge case situation.. if you click the chip and then drag and release over the X the button stays active

Fixed it too, I had a validation not working as expected.

docs/.vuepress/exampleComponents/ExampleChip.vue
juliodialpad: > must not be set when interactive is false

`isActive` is not set if the `interactive` prop is set to `false`

for the tabindex it should be -1 if `interactive` is `false` right?

braddialpad: since it's a span it's not focusable by default so it could just be unset. `tabindex=-1` would also work.

braddialpad (approved):
My comments have all been resolved, just the design issues that Natayi mentioned remaining @DestaDialpad

natayi-dialpad (approved):

PR #615, opened 2022-06-14
docs(toc): add table of contents

jonathan-paugh:
## Description

Adds a table of contents section to the right side of the content area in the documentation. This new section contains the subheadings that were previously on the left sidebar. The toc section will responsively nest itself inside the page contents below the header for mobile displays.

Desktop:
82987AD5-D543-4F4F-8F74-65544EF70F3E

Mobile:
image

## Pull Request Checklist


docs/.vuepress/config.ts
```diff
@@ -81,9 +83,11 @@ export default defineUserConfig({
 
   // markdown config
   markdown: {
-    toc: false,
     code: {
       lineNumbers: false
+    },
+    extractHeaders: {
+      level: [2] // Generated data header levels (used for toc)
     }
```
jonathan-paugh, line 94: Toc header depth can be set here, currently only renders level 2 sub-headers that were in the sidebar.

docs/.vuepress/theme/layouts/Layout.vue
```diff
@@ -1,25 +1,101 @@
...
-          <h1 v-if="$page.frontmatter.title" class="d-headline42 d-lh6 d-mb8 d-pt16" v-text="$page.frontmatter.title" />
-          <p v-if="$page.frontmatter.desc" class="d-fs20 d-lh6" v-html="$page.frontmatter.desc"></p>
-        </header>
-        <div v-if="$page.path.startsWith('/components')" class="d-d-flex d-jc-space-between">
-          <h2 class="d-fs16">Preview</h2>
-          <a v-if="$page.frontmatter.storybook_url" class="d-link d-fs14"
-             :href="$page.frontmatter.storybook_url" target="_blank">
-            <span class="d-mr4">View Storybook</span>
-            <icon-launch class="d-svg--size16" />
-          </a>
```
jonathan-paugh: I exported this header section out into its own component: `PageHeader.vue`

docs/.vuepress/theme/layouts/Layout.vue
```diff
@@ -1,25 +1,101 @@
...
+                    <PageToc v-if="isMobile" :headers="$page.headers" />
+                  </template>
+                </PageHeader>
+              </template>
+            </Page>
+          </Transition>
+          <PageToc v-if="!isMobile" :headers="$page.headers" class="d-pr32 d-pt24" />
         </div>
-      </section>
     </template>
```
jonathan-paugh, line 24: The main layout structure was modified to allow the toc to work with the page content.

Instead of inserting the header into the `#page-top` slot, the `#page` slot is now overridden.
This allows wrapping the `Page` and `PageToc` components in a flexbox so they display side-by-side.

I made sure to follow a similar structure that is used in the [default layout](https://github.com/vuepress/vuepress-next/blob/eb962a057a91142a46a071896de418bc9903c388/packages/@vuepress/theme-default/src/client/layouts/Layout.vue) to preserve original functionality such as transitions.

docs/.vuepress/theme/layouts/Layout.vue
```diff
@@ -1,25 +1,101 @@
...
+    align-self: flex-start;
+    background-color: var(--base--background-color);
+    z-index: 10;
+  }
+
+  @media screen and (max-width: 980px) {
+    .main-content {
+      flex-wrap: wrap;
+      flex-direction: column-reverse;
+    }
```
jonathan-paugh, line 75: I ended up using a breakpoint instead of the content width break implementation so the location of the `Toc` component in DOM can also be moved at this breakpoint. Also, this will break at the same exact point at the sidebar (980px).

docs/.vuepress/theme/index.js
```diff
@@ -1,23 +1,25 @@
-const { path } = require('@vuepress/utils')
-const { defaultTheme } = require('vuepress')
-const { mediumZoomPlugin } = require('@vuepress/plugin-medium-zoom')
+const {path} = require('@vuepress/utils')
+const {defaultTheme} = require('vuepress')
+const {mediumZoomPlugin} = require('@vuepress/plugin-medium-zoom')
```
josedialpad: Looks like we don't have eslint setup in Dialtone. I think we can add a similar configuration for dialtone as [we have set up in dialtone-vue](https://github.com/dialpad/dialtone-vue/blob/staging/.eslintrc.js) to prevent these type of changes (but it's not needed to add in this PR).

docs/.vuepress/theme/components/PageToc.vue
```diff
@@ -0,0 +1,25 @@
+<template>
```
josedialpad, line 1: Seems that we lost item highlighting on navigation? Also, the navigation of the sections was being reflected in the URL, I'm not sure if this can be configured in the plugin, it'd be nice to preserve it.

Example:
2022-06-14 at 11 57 53@2x

docs/.vuepress/theme/components/PageToc.vue
```diff
@@ -0,0 +1,25 @@
...
+  const options = {
+    containerTag: "",
+    listClass: "toc-list",
+    itemClass: "toc-item",
+    linkClass: "toc-link"
+  };
+</script>
+
+<style lang="less">
+  .toc .toc-list {
```
josedialpad: The Dialtone class `d-ls-reset` can be used here to reset the default list style https://dialpad.design/utilities/typography/lists.html#resetting-a-list

docs/.vuepress/theme/components/PageToc.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+  <aside v-if="headers.length > 0" class="toc">
+    <h3 class="d-pl24 d-pb8 d-ws-nowrap">ON THIS PAGE</h3>
```
josedialpad: nit: add class `d-tt-uppercase` and use just the normal text: `On this page`. Not a big deal in this case, but here are [some reasons why doing this](https://medium.com/@mandy.michael/why-you-shouldnt-write-your-content-in-uppercase-instead-use-css-b03ac2c65b99).

josedialpad (commented):
Hey Jonathan, I'm leaving a couple of comments but overall great first contribution 🙌

docs/.vuepress/theme/components/PageToc.vue
jonathan-paugh: I will look into this, I think it was working at some point. It might have to do with the changes to the header levels.

braddialpad: Yeah I remember seeing this working also, probably some type of customization that broke it

docs/.vuepress/theme/index.js
braddialpad: @josedialpad could you add a task for this in jira? Much more important now that we have vue code in the dialtone repo

josedialpad: > Much more important now that we have vue code in the dialtone repo

I agree. Created https://dialpad.atlassian.net/browse/DT-509

docs/.vuepress/theme/components/PageToc.vue
jonathan-paugh: It seems that the problem is caused by not rendering the sub-headers in the sidebar. A hacky solution would be to use `display: none` on nested `ul` tags in the sidebar. I just found this [active-header-links plugin](https://v2.vuepress.vuejs.org/reference/plugin/active-header-links.html) which looks promising, so I will try it out first.

braddialpad: I think that's actually a pretty decent solution if it works.

jonathan-paugh: The plugin worked like a charm, should be all good now.

jonathan-paugh:
Implemented all the mentioned changes.

braddialpad:
Super minor, but some of the links go from unwrapped to wrapped after selecting them such as "clipping text" here.

Screen Shot 2022-06-14 at 12 03 54 PM
Screen Shot 2022-06-14 at 12 04 08 PM

josedialpad:
Another thing that I don't know if it's related to what Brad mentioned is that the highlight on different items is causing a small shift in the main content for example in the [breadcrumbs page](https://dialpad.design/deploy-previews/pr-615/components/breadcrumbs.html), clicking on "Usage" and then in "Variants and Examples" causes a left shift, then click on "Classes" and the content goes to the right again.

josedialpad:
> ## Obligatory GIF (super important!)
> 

lol, I haven't seen the gif before 😂

braddialpad:
> lol, I haven't seen the gif before 😂

accurate, lol

braddialpad:
> clicking on "Usage" and then in "Variants and Examples" causes a left shift, then click on "Classes" and the content goes to the right again.

Yeah seems the whole right bar widens when you click variants and examples. Clicking through the rest of the others are fine. Maybe only happens on multi line items?

jonathan-paugh:
It seems to be only multi line headers that do that, seems to be okay if I use `d-ws-nowrap` on the item class. However I'm not sure if wrapping is something we want to keep. Imo it is probably better to have no wrapping, but long headers have the possibility of significantly reducing the content section size.

braddialpad:
> It seems to be only multi line headers that do that, seems to be okay if I use `d-ws-nowrap` on the item class. However I'm not sure if wrapping is something we want to keep. Imo it is probably better to have no wrapping, but long headers have the possibility of significantly reducing the content section size.

Can we set the TOC sidebar to a fixed width?

jonathan-paugh:
> Can we set the TOC sidebar to a fixed width?

Turns out it was cause by `font-weight` increasing the font size when the item is active, should be fixed now. Having a fixed size for the sidebar is possible, I don't know if it's needed now that this is working.

braddialpad:
> Turns out it was cause by `font-weight` increasing the font size when the item is active, should be fixed now. Having a fixed size for the sidebar is possible, I don't know if it's needed now that this is working.

Nope this is all good, good find.

braddialpad (approved):
Lets wait until Julio's vuepress styling fix is in, merge that into this branch and make sure everything still works before merging.

josedialpad (approved):

braddialpad:
Julio's fix is in

braddialpad:
Looks like a big gap at the top of the page since the merge. Not sure if this was due to your changes or Julios.
Screen Shot 2022-06-14 at 2 31 00 PM

jonathan-paugh:
Yeah, just fixed that. I added padding to the top of the header but it seems like it is not needed anymore. I'm going to look around a bit more to see if there are any more styling problems.

PR #616, opened 2022-06-14
fix(Chip): improvements to semantic html

braddialpad:
## Description
Updated the chip component to use more semantic html as well as various other improvements:

- Chip will now be a `<button>` element when it is interactive and a `<span>` element when it is not.
- Removed `d-chip--interactive` class as now the element type determines whether it is interactive or not (button vs span)
- Removed tabindex
- Close button is no longer nested within the chip itself, it is now within a wrapper containing the chip and the button. Absolute positioning is used to position the close button within the chip.
- Created `d-chip-close-btn` class.
- Removed `d-chip-btn-holder` class, it is now automatically set via a pseudo class.
- Use pseudo class to perform "oversized clickable area" on close button.

## Pull Request Checklist

### DIFF
--- a/docs/components/chip.md
+++ b/docs/components/chip.md
@@ -21,10 +21,10 @@ The base chip should be the go-to chip for most of your needs.
 
 ```html
 <span class="d-chip-container">
-  <button class="d-chip">
+  <button class="d-chip" type="button">
     <span>Chip</span>
   </button>
-  <button class="d-chip-close-btn" aria-label="close">...</button>
+  <button class="d-chip-close-btn" type="button" aria-label="close"/>
 </span>
 ```
 
@@ -35,7 +35,7 @@ The base chip should be the go-to chip for most of your needs.
 
 ```html
 <span class="d-chip-container">
-  <button class="d-chip">
+  <button class="d-chip" type="button">
     <span>Chip</span>
   </button>
 </span>
@@ -48,7 +48,7 @@ The base chip should be the go-to chip for most of your needs.
 
 ```html
 <span class="d-chip-container">
-  <button class="d-chip">
+  <button class="d-chip" type="button">
     <span class="d-chip__icon">...</span>
     <span>Chip</span>
   </button>
@@ -62,11 +62,11 @@ The base chip should be the go-to chip for most of your needs.
 
 ```html
 <span class="d-chip-container">
-  <button class="d-chip">
+  <button class="d-chip" type="button">
     <span class="d-chip__icon">...</span>
     <span>Chip</span>
   </button>
-  <button class="d-chip-close-btn" aria-label="close">...</button>
+  <button class="d-chip-close-btn" type="button" aria-label="close"/>
 </span>
 ```
 
@@ -77,11 +77,11 @@ The base chip should be the go-to chip for most of your needs.
 
 ```html
 <span class="d-chip-container">
-  <button class="d-chip">
+  <button class="d-chip" type="button">
     <span class="d-avatar">...</span>
     <span>...</span>
   </button>
-  <button class="d-chip-close-btn" aria-label="close">...</button>
+  <button class="d-chip-close-btn" type="button" aria-label="close"/>
 </span>
 ```
 
@@ -98,7 +98,7 @@ the close button can still be interactive even if the chip is non-interactive.
   <span class="d-chip">
     <span>Chip</span>
   </span>
-   <button class="d-chip-close-btn" aria-label="close">...</button>
+  <button class="d-chip-close-btn" aria-label="close"/>
 </span>
 ```
 
@@ -111,10 +111,10 @@ To truncate text, add `.d-truncate` to the content element, and set the width of
 
 ```html
 <span class="d-chip-container">
-  <button class="d-chip d-w102">
+  <button class="d-chip d-w102" type="button">
     <span class="d-truncate">Chip loooooong name</span>
   </button>
-  <button class="d-chip-close-btn" aria-label="close">...</button>
+  <button class="d-chip-close-btn" type="button" aria-label="close"/>
 </span>
 ```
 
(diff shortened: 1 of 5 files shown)


docs/components/chip.md
```diff
@@ -20,9 +20,11 @@ The base chip should be the go-to chip for most of your needs.
 </code-well-header>
 
 ```html
-<span class="d-chip d-chip--interactive" tabindex="0">
-  <span>...</span>
-  <button class="d-btn d-btn--circle" aria-label="close" type="button" >...</button>
+<span class="d-chip-container">
+  <button class="d-chip">
```
francisrupert: Missing `type` attribute.

francisrupert: Should probably also work as an `<a>` element, even if we're not showing an example of an `<a>`. I quickly modified it in Inspector and it appears to hold up well, except underline was still prsent.

docs/components/chip.md
```diff
@@ -20,9 +20,11 @@ The base chip should be the go-to chip for most of your needs.
...
+    <span>Chip</span>
+  </button>
+  <button class="d-chip-close-btn" aria-label="close">...</button>
```
francisrupert: Remove `...`. No need for it given `aria-label`.

docs/components/chip.md
```diff
@@ -20,9 +20,11 @@ The base chip should be the go-to chip for most of your needs.
 </code-well-header>
 
 ```html
-<span class="d-chip d-chip--interactive" tabindex="0">
-  <span>...</span>
-  <button class="d-btn d-btn--circle" aria-label="close" type="button" >...</button>
+<span class="d-chip-container">
```
francisrupert: We haven't discussed HTML Code Style yet, but this naming convention is against the grain from what I've seen (assuming our aim is to be more strictly BEM.

e.g. instead of...
```html
<span class="d-chip-container">
  <button class="d-chip">
    <span>Chip</span>
  </button>
  <button class="d-chip-close-btn" aria-label="close">...</button>
</span>
```
...do...
```html
<span class="d-chip">
  <button class="d-chip__label" type="button">
    <span class="d-chip__text">Chip</span>
  </button>
  <button class="d-chip__close" aria-label="close" type="button"></button>
</span>
```

docs/components/chip.md
```diff
@@ -66,38 +76,45 @@ The base chip should be the go-to chip for most of your needs.
 </code-well-header>
 
 ```html
-<span class="d-chip d-chip--interactive" tabindex="0">
-  <span class="d-avatar">...</span>
-  <span>...</span>
-  <button class="d-btn d-btn--circle" type="button">...</button>
+<span class="d-chip-container">
+  <button class="d-chip">
+    <span class="d-avatar">...</span>
```
francisrupert, line 93: Another BEM convention thing I highly recommend when nesting a component within a component:

Instead of...

```html
<span class="d-avatar">...</span>
````

...do...

```html
<span class="d-avatar d-chip__avatar">...</span>
````

That way we keep the styles less specific. So intead of...

```less
.d-chip {
  ...
  .d-avatar {...} // over-specific
}
```
...do...
```less
.d-chip {
  ...
  &__avatar {...} // flat
}
```

docs/components/chip.md
braddialpad: right, I thought the default type was `button` but it's not it's `submit`

braddialpad: added support for `<a>`

docs/components/chip.md
braddialpad: This actually needs to have an SVG nested inside it to display the icon. This isn't very obvious from this example so I modified it a bit

docs/components/chip.md
braddialpad: Yeah this is much nicer, thanks!

docs/components/chip.md
braddialpad: The nested `d-avatar` in this case has different properties for each chip size. Would you say it's still preferable to do it this way and have 3 separate classes `d-chip__avatar--xs`, `d-chip__avatar--sm`, `d-chip__avatar`? 

In the current nested implementation the avatar size is defined by the chip size, which does give some level of convenience.

francisrupert: Yeah, I'm ok with that convenience of Avatar size contextually to the `__label` size. I generally keep to the mantra of _flattest possible selector_ — but I'm not too dogmatic about it.

lib/build/less/components/chip.less
```diff
@@ -73,32 +57,42 @@
...
+  }
 }
 
-//  ============================================================================
-//  $   INTERACTIVITY
-//  ----------------------------------------------------------------------------
-//  $$  INTERACTIVE (DEFAULT)
-//  ----------------------------------------------------------------------------
-.d-chip--interactive {
+a.d-chip__label {
```
francisrupert: May as well combine these...

```less
button.d-chip__label,
a.d-chip__label {
  ...
}
```

braddialpad: thanks, done

francisrupert: From what I understand of our browser support targets, it'd be safe to use `:is()`, https://caniuse.com/?search=%3Ais. 

Specificity difference via https://specificity.keegan.st/
image

Assuming you agree, then,..
```less
button.d-chip__label, 
a.d-chip__label { 
  ... 
} 
```
...can be...

```less
.d-chip__label {

  ...
  
  &:is(a),
  &:is(button),
  &:is([role="button"]),
  &:is([role="link"]) {

    ...

  }

  ...

}
```

francisrupert: @braddialpad,

Other than this last review (and previously resolved ones), this should be good to go.

braddialpad: done, thanks!

PR #620, opened 2022-06-22
docs(vue live): add vue-live examples

juliodialpad:
## Description

Added [vue-live](https://github.com/vue-styleguidist/vue-live) so we can display dialtone-vue components and modify it live.

## Pull Request Checklist


juliodialpad:
Looks like we are not going to be able to use dialtone-vue components in dialtone as VuePress build is using SSR and dialtone-vue is not SSR ready... we are going to need to check how to make dialtone-vue SSR compatible.

braddialpad:
> Looks like we are not going to be able to use dialtone-vue components in dialtone as VuePress build is using SSR and dialtone-vue is not SSR ready... we are going to need to check how to make dialtone-vue SSR compatible.

Can we use ClientOnly? https://v2.vuepress.vuejs.org/reference/components.html

.github/workflows/deploy-preview.yml
```diff
@@ -33,6 +33,9 @@ jobs:
           **/node_modules/
         key: ${{ format('{0}-dialtone-node-modules-{1}-{2}', runner.os, env.DIALTONE_DEPLOY_PREVIEW_CACHE_VERSION, hashFiles('**/package-lock.json')) }}
 
+    - name: Authenticate with private NPM package
+      run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_AUTH_TOKEN }}" > ~/.npmrc
```
braddialpad, line 37: Nice, thanks. I think we will also have to add this to any other job that does an `npm install` right?

docs/.vuepress/baseComponents/CustomLayout.vue
```diff
@@ -0,0 +1,53 @@
+<template>
+  <div class="preview-code">
+    <div>
+      <slot name="editor"></slot>
+    </div>
+    <div>
+      <slot name="preview"></slot>
+    </div>
+  </div>
```
braddialpad: Not sure how "complete" this was intended to be since this is just a POC, but might as well define some of the styling of the custom layout now if we can to make future stories easier.

Seems to be an issue with the inner text color

Screen Shot 2022-06-23 at 9 36 31 AM

braddialpad: Is it possible to put example on top, code below?

braddialpad: Can the example have the same styling as the example above? (background color and such)

Screen Shot 2022-06-23 at 9 49 55 AM

docs/components/button.md
```diff
@@ -10,6 +10,15 @@ storybook_url: https://vue.dialpad.design/?path=/story/components-button--defaul
...
 </code-well-header>
 
+### Live preview
+
+<vue-live
+  class="d-bgc-black-600"
+  code="<button class='d-btn d-btn--primary'>Button</button>"
+  :editorProps="{ lineNumbers: true }"
+  :layout="CustomLayout"
+/>
```
braddialpad: We should wrap this in a "live-preview" component where you only pass in "code" and we'll probably need "components" too. Rest of the props here would be the defaults

braddialpad: What do you think about registering all dialtone-vue components in this live-preview component so they could be used on any example?

braddialpad: We may also need to include icons.. We should probably just include a few. registering every single icon might be a bit ridiculous

.github/workflows/deploy-preview.yml
juliodialpad: Right

docs/.vuepress/baseComponents/CustomLayout.vue
juliodialpad: > Is it possible to put example on top, code below?

Sure, I'll give it the same styling as our code-previews

docs/components/button.md
juliodialpad: > We should wrap this in a "live-preview" component where you only pass in "code" and we'll probably need "components" too. Rest of the props here would be the defaults

That's a good idea, I'll do it

juliodialpad:
> > Looks like we are not going to be able to use dialtone-vue components in dialtone as VuePress build is using SSR and dialtone-vue is not SSR ready... we are going to need to check how to make dialtone-vue SSR compatible.
> 
> Can we use ClientOnly? https://v2.vuepress.vuejs.org/reference/components.html

That one didn't work... found a solution in v1 documentation, seems like they haven't migrated that yet or forgot to do it.

https://vuepress.vuejs.org/guide/using-vue.html#browser-api-access-restrictions

Now it's working!

docs/components/button.md
juliodialpad: > We may also need to include icons.. We should probably just include a few. registering every single icon might be a bit ridiculous

We can import icons globally in `client.js` those components are available to vue-live too.

juliodialpad: > What do you think about registering all dialtone-vue components in this live-preview component so they could be used on any example?

I've registered dialtone-vue components globally so we can now use it on any place we want.

braddialpad: sweet

braddialpad:
> That one didn't work... found a solution in v1 documentation, seems like they haven't migrated that yet or forgot to do it.
> 
> https://vuepress.vuejs.org/guide/using-vue.html#browser-api-access-restrictions
> 
> Now it's working!

Nice good find

braddialpad (approved):
Really nice work on this, this is awesome! You can replace the original preview on the button page now with this one and you're good to merge!

PR #622, opened 2022-06-28
fix: dt-547 disable 'calt' font feature setting

braddialpad:
## Description
Bit of an odd "intentional feature" of the Inter font was causing some issues with text alignment within inputs. Original report here: https://dialpad.atlassian.net/browse/DP-51979

Confirmation that this is an intentional feature of the font: https://github.com/rsms/inter/issues/243

Disabled by setting the "calt" font-feature-setting on the body to 0, same place we set the font family.

Additionally, fonts were not loading in vuepress locally, so I moved them to the correct directory.

31d973ca-63f0-4977-927f-3247fad24d36

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@
 lib/dist
 docs/assets/css
 docs/assets/fonts
+docs/.vuepress/public/fonts
 docs/_includes/icons
 docs/_includes/patterns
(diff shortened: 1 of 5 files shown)


josedialpad (approved):
Looks like the majority of changes are coming from the linter

PR #623, opened 2022-06-30
docs: fix anchor scroll position

jonathan-paugh:
## Description
Add an offset when the window scrolls to an anchor position to compensate for sticky header size.

## Pull Request Checklist

### DIFF
--- a/docs/.vuepress/client.js
+++ b/docs/.vuepress/client.js
@@ -41,17 +41,6 @@ export default defineClientConfig({
       });
     }
 
-    const scrollBehavior = router.options.scrollBehavior;
-    router.options.scrollBehavior = async (to, from, savedPosition) => {
-      const anchorPadding = 16;
-      const position = await scrollBehavior(to, from, savedPosition);
-      return (
-        to.hash
-        ? { ...position, top: app.config.globalProperties.$headerSize + anchorPadding }
-        : position
-      );
-    }
-
     // Common views
     app.component('icons', Icons);
     app.component('colors', Colors);
--- a/docs/.vuepress/client.js
+++ b/docs/.vuepress/client.js
@@ -52,10 +52,6 @@ export default defineClientConfig({
       );
     }
 
-    window.addEventListener('hashchange', () => {
-      router.push(router.currentRoute.value);
-    });
-
     // Common views
     app.component('icons', Icons);
     app.component('colors', Colors);
--- a/docs/.vuepress/client.js
+++ b/docs/.vuepress/client.js
@@ -41,6 +41,17 @@ export default defineClientConfig({
       });
     }
 
+    const scrollBehavior = router.options.scrollBehavior;
+    router.options.scrollBehavior = async (to, from, savedPosition) => {
+      const anchorPadding = 16;
+      const position = await scrollBehavior(to, from, savedPosition);
+      return (
+        to.hash
+        ? { ...position, top: app.config.globalProperties.$headerSize + anchorPadding }
+        : position
+      );
+    }
+
     // Common views
     app.component('icons', Icons);
     app.component('colors', Colors);
--- a/docs/.vuepress/client.js
+++ b/docs/.vuepress/client.js
@@ -52,6 +52,10 @@ export default defineClientConfig({
       );
     }
 
+    window.addEventListener('hashchange', () => {
+      router.push(router.currentRoute.value);
+    });
+
     // Common views
     app.component('icons', Icons);
     app.component('colors', Colors);
--- a/docs/.vuepress/theme/layouts/Layout.vue
+++ b/docs/.vuepress/theme/layouts/Layout.vue
@@ -56,6 +56,7 @@
       ? mobileHeaderSize
       : defaultHeaderSize
     );
+    console.log(app.appContext.config.globalProperties.$headerSize);
   }, { immediate: true });
 </script>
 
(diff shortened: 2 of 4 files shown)


docs/.vuepress/client.js
```diff
@@ -41,6 +41,17 @@ export default defineClientConfig({
...
+    const scrollBehavior = router.options.scrollBehavior;
+    router.options.scrollBehavior = async (to, from, savedPosition) => {
+      const anchorPadding = 16;
+      const position = await scrollBehavior(to, from, savedPosition);
+      return (
+        to.hash
+        ? { ...position, top: app.config.globalProperties.$headerSize + anchorPadding }
+        : position
+      );
+    }
```
jonathan-paugh: VuePress creates the router and then [provides the instance](https://v2.vuepress.vuejs.org/advanced/cookbook/usage-of-client-config.html#enhance) so we don't have access to set the [scrollBehavior](https://router.vuejs.org/api/#scrollbehavior) with `createRouter` as intended. Instead it seems the best solution is to wrap the function provided on the router instance and modify the offset if it is an anchor.

docs/.vuepress/theme/layouts/Layout.vue
```diff
@@ -32,20 +32,31 @@
...
 
-  const scrollPromise = useScrollPromise();
-  const onBeforeEnter = scrollPromise.resolve;
-  const onBeforeLeave = scrollPromise.pending;
+  watch(isMobile, (current) => {
+    app.appContext.config.globalProperties.$headerSize = (
+      current
+      ? mobileHeaderSize
+      : defaultHeaderSize
+    );
```
jonathan-paugh: `globalProperties` may not be the best way to share this value. [Provide / Inject](https://vuejs.org/guide/components/provide-inject.html#provide) seemed like a promising solution, however only `Provide(...)` can be used at both the app and component level but `Inject(...)` can only be used at the component level and it is needed at the app level.

josedialpad:
2022-06-30 at 17 47 52@2x

Is it possible to apply this fix to the header anchor? Seems that's the same issue with the TOC links

jonathan-paugh:
> ...
> Is it possible to apply this fix to the header anchor? Seems that's the same issue with the TOC links

Hmm I didn't notice that, I'll look into it.

braddialpad (changes requested):
While this works for the most part, it's a lot of weird overrides and pretty hard to understand. I think it might be possible to implement a pure css solution using [scroll-margin-top](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin-top) `scroll-margin-top: [OFFSET]px;`. This would have to be generated to be on each heading, but I locally edited the CSS and it seemed to work.

Let's at least try this, if it doesn't work for some reason we can go back to the method in this PR.

docs/.vuepress/theme/layouts/Layout.vue
josedialpad: Another approach that could be simpler, is to use the `--navbar-height` custom property that vuepress has. Actually, I checked in the [Vuepress docs](https://v2.vuepress.vuejs.org/guide/#why-not) and seems that using a combination of 
```
margin-top: calc(0.5rem - var(--navbar-height));
padding-top: calc(1rem + var(--navbar-height));
```
does the trick for all the links.

2022-06-30 at 18 23 35@2x

jonathan-paugh: It seems we do have this styling:

image

It gets overridden by a bunch of other stylings:

image

If I manually force it to take on the margin and padding it seems to work well.

I found the main style that is overriding it in `docs/assets/css/dialtone-docs.css` which seems to be an auto generated file, ~~I'm not sure where it comes from.~~ It comes from the less file.

The other style overriding is from `docs/.vuepress/styles/palette.scss` which I can just remove, so it's not a problem.

docs/.vuepress/theme/layouts/Layout.vue
```diff
@@ -43,9 +49,13 @@
...
-  const scrollPromise = useScrollPromise();
-  const onBeforeEnter = scrollPromise.resolve;
-  const onBeforeLeave = scrollPromise.pending;
+  watch(isMobile, (current) => {
+    const root = document.querySelector(':root');
+    root.style.setProperty(
+      '--navbar-height',
+      `${(current ? mobileHeaderSize : defaultHeaderSize)}px`
+    );
+  });
```
jonathan-paugh: It now modifies the css variable for `navbar-height` instead of `globalProperty`.

docs/assets/less/overrides.less
```diff
@@ -123,7 +123,8 @@ samp {
 
   > h2,
   > h3 {
-    margin-top: var(--su48);
+    margin-top: calc(0.5rem - var(--navbar-height) + var(--su48));
+    padding-top: calc(1rem + var(--navbar-height));
```
jonathan-paugh, line 127: Reverted all the other code, this is all that was needed. 🙂

braddialpad (approved):
Muuuuuuch nicer!

docs/.vuepress/theme/layouts/Layout.vue
josedialpad: Just add the `immeadiate` option to the watcher so it sets the correct value from the first run

josedialpad (approved):
w/ a comment

PR #635, opened 2022-08-03
fix(modal): fix spacing and overriding styles in the docsite

josedialpad:
## Description
Fixes the overriding styles in the docsite.
It also fixes the spacing in the variant when no footer is provided.

## Pull Request Checklist

### DIFF
--- a/lib/build/less/components/modal.less
+++ b/lib/build/less/components/modal.less
@@ -96,8 +96,6 @@
   visibility: hidden;
   opacity: 0;
   will-change: visibility, z-index, opacity, transform;
-
-  .d-stack16();
 }
 
 //  $$   MAKE THEM APPEAR
@@ -139,10 +137,10 @@
 //  ----------------------------------------------------------------------------
 .d-modal__header {
   margin: 0 !important;
-  padding: var(--su24) var(--su24) 0;
+  padding: var(--su24) var(--su24) 0 !important;
   color: var(--modal-header--fc);
   font-weight: var(--fw-bold);
-  font-size: var(--fs24);
+  font-size: var(--fs24) !important;
   line-height: var(--lh4);
 }
 
@@ -150,7 +148,7 @@
 //  ----------------------------------------------------------------------------
 .d-modal__content {
   max-width: 75ch;
-  margin: var(--su12) 0 0 !important;
+  margin: var(--su12) 0 !important;
   padding: var(--su4) var(--su24);
 }
 
--- a/lib/build/less/components/modal.less
+++ b/lib/build/less/components/modal.less
@@ -137,10 +137,10 @@
 //  ----------------------------------------------------------------------------
 .d-modal__header {
   margin: 0 !important;
-  padding: var(--su24) var(--su24) 0 !important;
+  padding: var(--su24) var(--su24) 0;
   color: var(--modal-header--fc);
   font-weight: var(--fw-bold);
-  font-size: var(--fs24) !important;
+  font-size: var(--fs24);
   line-height: var(--lh4);
 }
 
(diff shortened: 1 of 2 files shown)


lib/build/less/components/modal.less
```diff
@@ -139,18 +137,18 @@
 //  ----------------------------------------------------------------------------
 .d-modal__header {
   margin: 0 !important;
-  padding: var(--su24) var(--su24) 0;
+  padding: var(--su24) var(--su24) 0 !important;
   color: var(--modal-header--fc);
   font-weight: var(--fw-bold);
-  font-size: var(--fs24);
+  font-size: var(--fs24) !important;
```
francisrupert: Is this a bit too brute-force? I thought the real culprit was VuePress' over-specific CSS selectors leaking into our components? 

Can VuePress' CSS be updated rather than updating the DT component itself?

The age-old "avoid `!important`" is the go-to thinking here, particularly since we're deliberate about `!important` for Utilities.

josedialpad: Yeah I agree and the first approach I've tried was to adapt the styles not to override the modal's style, and I found that we're overriding the styles here https://github.com/dialpad/dialtone/blob/staging/docs/.vuepress/styles/palette.scss#L11-L24 that's in our side and can be modified, but also the Vuepress default theme (which we're using) has a set of styles we cannot modify since they're coming from the `@vuepress` package https://github.com/vuepress/vuepress-next/blob/main/packages/%40vuepress/theme-default/src/client/styles/layout.scss#L123-L152

What I found in Vuepress v1 is the `eject` command to copy the default theme source into the project so that way we can customize it but that implies the theme will stop to receive any update and will require manual update. Also the `eject` command was removed in the v2 https://v2.vuepress.vuejs.org/guide/migration.html#cli-change:~:text=%23-,eject%20command,-Removed.

braddialpad: Is there a way we can change the specificity of the CSS in this component to make this work without important? 

Additionally, looks like `.theme-default-content` is the problematic class, is there a way we can exclude `.theme-default-content` from affecting our examples at the Vuepress level? Is it possible to use `:not()` or something like that?

josedialpad: > Additionally, looks like .theme-default-content is the problematic class, is there a way we can exclude .theme-default-content from affecting our examples at the Vuepress level? Is it possible to use :not() or something like that?

I couldn't find an easy way, to do it it requires changing the layout or extending the default theme.

> Is there a way we can change the specificity of the CSS in this component to make this work without important?

Updated to fix with specificity https://github.com/dialpad/dialtone/pull/635/commits/9aed7077a8e14cd4e973b7a288b32e4211423dc1

Basically, we need to override the `.d-modal__header` class just giving higher specificity to win the battle against the styles coming from the default theme that's not possible to modify https://github.com/vuepress/vuepress-next/blob/main/packages/%40vuepress/theme-default/src/client/styles/layout.scss#L123-L152

francisrupert: Quite a workaround, nicely done. While it's a blatant code smell...

image

...it's as reasonable a workaround as any, better than `!important`. Hopefully we won't have to do this too often going forward, let alone if we haven't caught other instances of VuePress styles leaking in.

Very unfortunate VuePress didn't realize how much their style approach _will_ leak in. A significant redesign/rearchitecture of doc site is on the backburner, but nowhere near a priority. 

I approve this change, what say you @braddialpad?

braddialpad: Yep I'm okay with this.

braddialpad (approved):

PR #636, opened 2022-08-08
docs: add migration guide

juliodialpad:
## Description

Added migration guide in Markdown format, we still need to decide if we want the guide to be on the docsite or in confluence internally.

Some eslint errors came out and decided include the fix here too.

## Pull Request Checklist


migration_guide/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,35 @@
...
+   - (purple-200) `#D1C0FF` → `#D3BCFF` (purple-200)
+   - (purple-300) `#A687FF` → `#AB7EFF` (purple-300)
+   - (purple-400) `#8B65FF` → `#7C52FF` (purple-400)
+   - (purple-500) `#6C3DFF` → `#7C52FF` (purple-400)
+   - (purple-600) `#3A1C95` → `#3A1D95` (purple-500)
+   - (purple-700) `#2A1173` → `#3A1D95` (purple-500)
+   - (purple-800) `#160645` → `#10022C` (purple-600)
+4. At last, check the usages you just replaced and ensure you still have the desired look and feel,
+contrast radio. If not, check with your designer which background/foreground color you should update 
+to meet the accessibility requirements.
```
braddialpad: Should we also mention to update dialtone-vue and all the components should be immediately updated with the new colors?

migration_guide/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,35 @@
...
+- Merged old `purple-600` and `purple-700` stops into new `purple-500` stop
+- Mapped old `purple-800` stop into new `purple-600` stop
+
+![](purple_mappings.png)
+![](previous_next_mappings.png)
+
+## Migrating to v7
+
+1. Install dialtone@7.X
+   - `npm install @dialpad/dialtone@7`
```
francisrupert: Are we having them install the **Beta** version ([`v7.0.0-beta.2`](https://github.com/dialpad/dialtone/releases/tag/v7.0.0-beta.2)), or an as-yet NPM-published `7.x.x`?

braddialpad (commented):
Nice job, just one comment.

migration_guide/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,35 @@
...
+   - (purple-100) `#ECE1FF` → `#EEE5FF` (purple-100)
+   - (purple-200) `#D1C0FF` → `#D3BCFF` (purple-200)
+   - (purple-300) `#A687FF` → `#AB7EFF` (purple-300)
+   - (purple-400) `#8B65FF` → `#7C52FF` (purple-400)
+   - (purple-500) `#6C3DFF` → `#7C52FF` (purple-400)
+   - (purple-600) `#3A1C95` → `#3A1D95` (purple-500)
+   - (purple-700) `#2A1173` → `#3A1D95` (purple-500)
+   - (purple-800) `#160645` → `#10022C` (purple-600)
+4. At last, check the usages you just replaced and ensure you still have the desired look and feel,
+contrast radio. If not, check with your designer which background/foreground color you should update
```
francisrupert: "raDio" = "raTio"

migration_guide/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,35 @@
+# Dialtone 7 Migration guide
+
+The migration from v6 to v7 is mostly visual. The purple color stops were updated.
```
francisrupert, line 3: Would be good to note that while most of the migration is shouldered by these Dialtone updates, they must also search through their codebases as outlined below align.

migration_guide/MIGRATION_GUIDE.md
juliodialpad: Yeah, forgot about that, I'll add it as a note for 'dialtone-vue' users.

migration_guide/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,35 @@
...
+3. Replace any old RAW HEX values with the updated one and if you can replace RAW 
+values with the corresponding CSS variable, please do it.
+   - (purple-100) `#ECE1FF` → `#EEE5FF` (purple-100)
+   - (purple-200) `#D1C0FF` → `#D3BCFF` (purple-200)
+   - (purple-300) `#A687FF` → `#AB7EFF` (purple-300)
+   - (purple-400) `#8B65FF` → `#7C52FF` (purple-400)
+   - (purple-500) `#6C3DFF` → `#7C52FF` (purple-400)
+   - (purple-600) `#3A1C95` → `#3A1D95` (purple-500)
+   - (purple-700) `#2A1173` → `#3A1D95` (purple-500)
+   - (purple-800) `#160645` → `#10022C` (purple-600)
```
francisrupert, line 36: I'm curious: are there instances where products are doing this? I'm assuming consuming products mostly reference the CSS Custom Properties (eg. `var(--purple-300)` or use CSS Utilities), and (hopefully not much or at all) actual hard-coded HEX values (let alone RGB, HSL, etc).

migration_guide/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,35 @@
...
+- Mapped old `purple-800` stop into new `purple-600` stop
+
+![](purple_mappings.png)
+![](previous_next_mappings.png)
+
+## Migrating to v7
+
+1. Install dialtone@7.X
+   - `npm install @dialpad/dialtone@7`
+2. Replace the purple stops (`@purple-X`, `--purple-X`) in this order.
```
francisrupert: Do we need to be more _specific_ of the strings to search-and-replace, or do we feel confident that the below strings cover both CSS Custom Properties (`var({color})`) CSS Utilities (e.g. `d-bgc-{color}`, `d-fc-{color}`, etc)?

migration_guide/MIGRATION_GUIDE.md
juliodialpad: Do we have any Dialtone-vue version released with dialtone7? I think currently we only have a branch, right?

migration_guide/MIGRATION_GUIDE.md
juliodialpad: I was thinking the same, but I found some RAW values on Dialtone so I think it's worth to mention it.

francisrupert: Yep, safe call. 

Meanwhile, if you found RAW values in either Dialtone or Dialtone Vue sounds like an issue should be logged to reconcile.

migration_guide/MIGRATION_GUIDE.md
juliodialpad: I feel confident that we have good naming convention so {color} is always going to be an instance of `--purple-X` no matter if it has some prefix or not.

I think we discouraged the usage of LESS vars (`@purple-X`), and using CSS vars instead (`var(--purple-X)`) but I've found a lot more usages of LESS than CSS vars in the product.

Should we include in the replacement of those on the guide too?

migration_guide/MIGRATION_GUIDE.md
braddialpad: Probably not. We can make a version7 branch and release it as beta like we did with dialtone.

migration_guide/MIGRATION_GUIDE.md
braddialpad: Yeah ideally everything uses css vars instead of less vars. Not sure that should really be included in the scope of this migration though.

francisrupert: It looks like the guide as authored covers the LESS variables (i.e. "Replace the purple stops (`@purple-X`, `--purple-X`) in this order.").

If you're asking if we should require them to replace something like...
```less
.foo { color: @purple-999 }
```
...with...
```less
.foo { color: var(--purple-999) }
```
...that sounds like a good idea. However... is that something that should be considered outside the scope of this really-quick-update-everyone-wants-done-yesterday and can thus be an issue to be logged and prioritized for another time?

If nothing else, sounds like there's a Dialtone-specific issue that should be logged to audit Dialtone code to replace LESS variables with CSS Custom Properties where relevant. 

Meanwhile.... it occurs to me that we might be using LESS variables as a fallback, and that's a good thing right? Example: `--asdf: (--foo-999, @foo-999)`.

migration_guide/MIGRATION_GUIDE.md
braddialpad: will be beta for now, when we actually want to go to production with it we'll remove the beta

migration_guide/MIGRATION_GUIDE.md
braddialpad: > ...that sounds like a good idea. However... is that something that should be considered outside the scope of this really-quick-update-everyone-wants-done-yesterday and can thus be an issue to be logged and prioritized for another time?

Yep, given the urgency there appears to be for this

> If nothing else, sounds like there's a Dialtone-specific issue that should be logged to audit Dialtone code to replace LESS variables with CSS Custom Properties where relevant.

It's fine for us to use them in Dialtone code itself, it's in the product where it's an issue. The reason is mostly performance related. Every time in dialpad when they do `@import "@dialpad/dialtone/variables";` it's a performance hit. This gets REALLY bad when there's thousands of files doing it.

> 
> Meanwhile.... it occurs to me that we might be using LESS variables as a fallback, and that's a good thing right? Example: `--asdf: (--foo-999, @foo-999)`.

They probably shouldn't be doing this in product due to above mentioned performance issues. If we want to do this in dialtone that's fine.

migration_guide/MIGRATION_GUIDE.md
juliodialpad: I mean, most of them were in SVG's and just a few in VuePress/Storybook configurations.

francisrupert (approved):

braddialpad (approved):

PR #645, opened 2022-08-22
docs: remove header links greater than L2

braddialpad:
## Description
Header link anchors at any level greater than 2 were not working. We only want to display L2 links by design, this change removes all > L2 link anchors.

### DIFF
--- a/docs/.vuepress/config.ts
+++ b/docs/.vuepress/config.ts
@@ -86,6 +86,9 @@ export default defineUserConfig({
 
   // markdown config
   markdown: {
+    anchor: {
+      level: [1, 2],
+    },
     code: {
       lineNumbers: false
     },


docs/.vuepress/config.ts
```diff
@@ -86,6 +86,9 @@ export default defineUserConfig({
 
   // markdown config
   markdown: {
+    anchor: {
+      level: [1, 2],
+    },
```
francisrupert, line 91: @braddialpad Ah, I see what's happening here. The plugin Jonathan mentioned collides with deep links, so your change here suppresses beyond level 2.

I'd consider this only a temporary solution to overcome that current collision. I'm good with this config for the moment, but would like to resurrect that deep-linkability. Perhaps a new task should be logged to arrive at a solution that allows for sharing deep-links while restricting TOC levels. Or... bring back visibility of all TOC levels.

jonathan-paugh: I agree that it would be best to have access to all the header levels, the TOC can have custom header data passed to it.
Just thinking about it briefly the best way to go about it would probably be:
1. Change `extractHeaders` to generate all the header levels.
2. Filter the header data going into the TOC manually
    - This logic can be done directly in the 'PageToc.vue' component as it already receives the `headers` prop
3. The markdown generator and active-header-links plugin should receive the unfiltered header data by default (...I think). This should allow all anchors to show in the url and be shared.

braddialpad: Merging this and creating a jira for the future work.

PR #655, opened 2022-09-01
docs: add .nojekyll file to output folder.

braddialpad:
Fixes an issue with github pages where it will not render files starting with _


jonathan-paugh (approved):
Thanks!

PR #684, opened 2022-09-20
fix: input icon to position outside of input without absolute position

gisellezeitoune:
## Description
Change input icon to be outside of input without absolute position so when extensions such as lastpass positions icons inside the input, they don't overlap with our icons

Screen Shot 2022-09-20 at 12 17 23

## Pull Request Checklist

### DIFF
--- a/docs/components/input.md
+++ b/docs/components/input.md
@@ -228,8 +228,8 @@ Provides feedback to the user based on their interaction, or lack thereof, with 
         <label class="d-label" for="Dialtone--InputExample--IconRight">Label</label>
       </div>
       <div class="d-input__wrapper">
-        <span class="d-input-icon d-input-icon--right"><icon-lock /></span>
         <input class="d-input d-input-icon--right" id="Dialtone--InputExample--IconRight" type="text" placeholder="Placeholder" />
+         <span class="d-input-icon d-input-icon--right"><icon-lock /></span>
       </div>
     </div>
   </div>
@@ -246,8 +246,8 @@ Provides feedback to the user based on their interaction, or lack thereof, with 
 <div>
   <label class="d-label" for="Dialtone--InputExample--IconRight">...</label>
   <div class="d-input__wrapper">
-    <span class="d-input-icon d-input-icon--right">...</span>
     <input class="d-input d-input-icon--right" id="Dialtone--InputExample--IconRight" type="text" placeholder="..." />
+     <span class="d-input-icon d-input-icon--right">...</span>
   </div>
 </div>
 ```
--- a/lib/build/less/components/input.less
+++ b/lib/build/less/components/input.less
@@ -18,19 +18,12 @@
 //    - Sizes
 //  • INPUT ICONS
 //
-//  ============================================================================
-//  Inputs CSS Vars
-//  ----------------------------------------------------------------------------
-@inputs-icon-spacing--xs: calc(var(--su4) + @icon12 + var(--su8));
-@inputs-icon-spacing--sm: calc(var(--su4) + @icon14 + var(--su8));
-@inputs-icon-spacing:     calc(var(--su4) + @icon16 + var(--su8));
-@inputs-icon-spacing--lg: calc(var(--su4) + @icon20 + var(--su8));
-@inputs-icon-spacing--xl: calc(var(--su4) + @icon24 + var(--su8));
 
 //  $  INPUTS
 //  ----------------------------------------------------------------------------
 .d-input,
-.d-textarea {
+.d-textarea,
+.d-input__wrapper {
     //  Component CSS Vars
     //  ------------------------------------------------------------------------
     --input--focus-bc: var(--primary-color);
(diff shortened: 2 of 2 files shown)


docs/components/input.md
```diff
@@ -228,8 +228,8 @@ Provides feedback to the user based on their interaction, or lack thereof, with
         <label class="d-label" for="Dialtone--InputExample--IconRight">Label</label>
       </div>
       <div class="d-input__wrapper">
-        <span class="d-input-icon d-input-icon--right"><icon-lock /></span>
         <input class="d-input d-input-icon--right" id="Dialtone--InputExample--IconRight" type="text" placeholder="Placeholder" />
+         <span class="d-input-icon d-input-icon--right"><icon-lock /></span>
```
braddialpad, line 232: I was thinking there may need to be a change to the dialtone vue component now that the order of the elements matter, but seems we are already putting the right icon after the input and the left before.

lib/build/less/components/input.less
```diff
@@ -107,57 +94,46 @@
...
+        border: none;
+        border-radius: var(--base--corner-radius);
+
+        &:focus {
+            border: none;
+            outline: 0;
+            box-shadow: none !important;
+        }
+    }
 }
```
braddialpad: A small difference between the old and the new. The original spacing between the icon element and the placeholder text was 5px. In the new one it is 8px, see comparison below. 

5px is a super weird value to use and isn't part of the design system so I'd be okay with you changing this to `var(--su6)` instead.

old
Screen Shot 2022-09-20 at 9 53 06 AM


new
Screen Shot 2022-09-20 at 9 53 15 AM

braddialpad (commented):
Nice job, this is much cleaner than how we were doing it previously.

docs/components/input.md
gisellezeitoune: Yes, I checked in dialtone-vue and it's already that way. Is there another place I should be checking for this?

braddialpad: Nope, you're good on this one.

lib/build/less/components/input.less
gisellezeitoune: Done, I changed it to var(--su6)

braddialpad: This has reduced the horizontal padding on the entire input to 6px. It needs to be just between the icon and the text. I think we may have to use a method similar to how we were setting padding previously

```less
.d-input-icon {

    // Update padding for d-input, d-textarea only if icon is present.
    &.d-input-icon--left ~ .d-input,
    &.d-input-icon--left ~ .d-textarea {
        padding-left: var(--su6);
    }
}
```

Unfortunately I don't think this will work for the right side anymore since we have moved the icon to be after the input in the dom so the ~ selector won't work.. I will keep trying to think of a solution...

lib/build/less/components/input.less
```diff
@@ -44,7 +37,7 @@
     box-sizing: border-box;
     width: 100%;
     min-width: 0;
-    padding: 0.7rem var(--su8);
+    padding: 0.7rem var(--su6);
```
braddialpad: can change this back to --su8 now

lib/build/less/components/input.less
braddialpad: Alright I think we can just put this back in within .d-input, .d-textarea

```less
.d-input,
.d-textarea {
  &.d-input-icon--right {
      padding-right: var(--su6);
  }
  
  &.d-input-icon--left {
      padding-left: var(--su6)
  }
}
```

docs/components/input.md
```diff
@@ -246,8 +246,8 @@ Provides feedback to the user based on their interaction, or lack thereof, with
 <div>
   <label class="d-label" for="Dialtone--InputExample--IconRight">...</label>
   <div class="d-input__wrapper">
-    <span class="d-input-icon d-input-icon--right">...</span>
     <input class="d-input d-input-icon--right" id="Dialtone--InputExample--IconRight" type="text" placeholder="..." />
+     <span class="d-input-icon d-input-icon--right">...</span>
```
braddialpad, line 250: It's possible this could break current inputs in the product not using dialtone vue components. I'll do a quick search here see if I can find any instances

braddialpad: I found one single instance in UC using `d-input-icon--right` that will have to be updated when we roll this out

lib/build/less/components/input.less
gisellezeitoune: It's fixed now

lib/build/less/components/input.less
gisellezeitoune: done

docs/components/input.md
gisellezeitoune: where's that? Should I be making the change?

braddialpad: hmm I guess I can just do it when we do the next release.

braddialpad (approved):
Looks good thanks!

PR #692, opened 2022-09-22
docs(components): add dialtone vue components

juliodialpad:
## Description

- Synced Dialtone-vue components with Dialtone documentation.
- Fixed a lot of broken links.
- Removed `Markdown` component due to it has no sense having a Markdown renderer component inside a Markdown file or Vue component, refactored usages.
- Improved overall site navigation using `RouterLink` instead of `<a>` tags

## Pull Request Checklist


docs/components/badge.md
```diff
@@ -15,18 +15,18 @@ storybook_url: https://vue.dialpad.design/?path=/story/components-badge--default
   <div class="dialtone-usage__item dialtone-usage__item--do">
     <h3 class="dialtone-usage__hd dialtone-usage__hd--do"><icon-checkmark /> Use for</h3>
     <div class="dialtone-usage__bd">
-<Markdown>
-- To flag and draw awareness to a specific element or feature of focus. For example, something is unique about that separates it from other like content.
-- As a notification system with minimal footprint.
-</Markdown>
```
braddialpad: Not sure we want to remove the markdown component, the point of the markdown component was so that we could render markdown syntax directly in the code like this. It makes it easier for people such as designers who don't know HTML to update the documentation, and easier to copy paste markdown directly into the docs.

braddialpad (changes requested):
The new component pages look good, but we'll need to add the markdown renderer back.

docs/components/badge.md
juliodialpad: I think this could be fixed in a better way, I could create a custom vue component for Do's and Don'ts that was the custom content that needed <Markdown>

I left those as HTML as it was easier to leave them like that to finish off the migration ASAP, but I can fix it correctly now so we can remove the <Markdown> component.

braddialpad: A specific do's and don'ts component might be of use since it's reused in so many places on the site, but regardless of that the point of this component was so we could pass markdown into components whenever we needed to. @francisrupert this was your ask originally I'm assuming you still would like to use it?

francisrupert: Definitely prefer to retain it, mostly on behalf of anyone that isn't us. This includes uses of `<Markdown>` that aren't directly part of the Do/Don't section. A custom Do/Don't totally makes sense, though that's not the only place `<Markdown>` would be useful.

For example, our Content Designer will soon be taking an active role in editing (including beyond Content Guidelines 😁) and 99% of that will be in Markdown. The custom `<Markdown>` element overcomes ability to author markdown beyond the root level of an `.md` page.... e.g.

```
<div>
- This
- Will not
- Be
- A
- List
</div>
```
...thus...
```
<div>
<Markdown>
- This
- *is*
- A
- List
- </Markdown>
</div>
```

francisrupert (commented):
This is great stuff, and goes a long way in reconciling b/t Dialtone and Dialtone Vue. I've logged [DT-715](https://dialpad.atlassian.net/browse/DT-715) to figure out how to make them more cohesive.

docs/components/badge.md
juliodialpad: Markdown can be put inside Vue components, for example:

```html
<div>

- This
- Will
- Be
- A
- List
</div>
```

You just need to put an space between the tag and the markdown and Vuepress will render it correctly.

https://v2.vuepress.vuejs.org/advanced/cookbook/markdown-and-vue-sfc.html

braddialpad: Alright so I just tried this, it DID work when I passed a list into a slot, but it didn't seem to work if I put it directly into an SFC. Does it work for you directly in an SFC? Also strange that the "newline to render markdown" thing doesn't really seem to be documented.

Tried this on the new dialtone usage component:

```vue
<!-- eslint-disable vue/html-indent -->
<!-- eslint-disable vue/multiline-html-element-content-newline -->
<template>
  <div class="dialtone-usage">
<div>

- This
- Will
- Be
- A
- List
</div>
    <div class="dialtone-usage__item dialtone-usage__item--do">
      <h3 class="dialtone-usage__hd dialtone-usage__hd--do">
        <icon-checkmark />
        Use for
      </h3>
      <div class="dialtone-usage__bd">
        <slot name="do" />
      </div>
    </div>
    <div class="dialtone-usage__item dialtone-usage__item--dont">
      <h3 class="dialtone-usage__hd dialtone-usage__hd--dont">
        <icon-close />
        Don't use for
      </h3>
      <div class="dialtone-usage__bd">
        <slot name="dont" />
      </div>
    </div>
  </div>
</template>

<script>
export default {
  name: 'DialtoneUsage',
};
</script>
```

And it rendered like this:
Screen Shot 2022-09-29 at 10 52 13 AM

braddialpad: Alright I THINK I understand what is going on here now. Only data from actual `.md` files gets rendered as markdown. So because we are passing markdown into the slot FROM a `.md` file, it converts it and passes it into the slot. If we render markdown directly in a `.vue` component, it doesn't get converted and renders as text. I think this should be fine as long as we design our vue components never to render markdown directly. Any markdown would always have to be passed in through a slot.

juliodialpad: Why would you like to put markdown into a SFC? we need markdown to render into slots so anyone can collaborate to documentation without them having the need to write code, right? 🤔

braddialpad: Maybe we keep the Markdown component in case we ever need to render Markdown within a vue component even though it's not currently being used?

juliodialpad: > Maybe we keep the Markdown component in case we ever need to render Markdown within a vue component even though it's not currently being used?

I cannot think of any use-case where we could need to render markdown inside our SFC, we can render markdown and Vue directly on `.md` files, I think that suites any use-case that could come up.

This is the documentation on using Vue and Markdown on `.md` files: https://v2.vuepress.vuejs.org/advanced/cookbook/markdown-and-vue-sfc.html

braddialpad: Yeah I think it's probably bad practice to render markdown directly in vue components. I'm okay with this.

braddialpad (approved):

PR #710, opened 2022-10-07
feat: add linux and android system fonts to the stack

braddialpad:
## Description
Apparently since we changed to our new system font stack it looks like this on Ubuntu:

Seems on linux sometimes none of our font stack exists and it was defaulting to joypixels in this case.

Included Ubuntu (Ubuntu system font), Cantarell (Gnome system font) and Roboto (Android system font) into the font stack.

## Pull Request Checklist

### DIFF
--- a/lib/build/less/variables/typography.less
+++ b/lib/build/less/variables/typography.less
@@ -19,8 +19,8 @@
 //  $   FONT FAMILY
 //  ----------------------------------------------------------------------------
 @define-ff-custom-name:                    -apple-system, BlinkMacSystemFont, 'Segoe UI Adjusted',  'Segoe UI', SFMono,
-                                           'Helvetica Neue', Arial, 'Noto Sans', sans-serif, 'Apple Color Emoji',
-                                           'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
+                                           'Helvetica Neue', Cantarell, Ubuntu, Roboto, Arial, 'Noto Sans', sans-serif,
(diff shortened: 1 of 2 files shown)


lib/build/less/variables/typography.less
```diff
@@ -66,7 +66,8 @@
 
 
 @ff-custom: @define-ff-custom-name;
-@ff-sans: -apple-system, BlinkMacSystemFont, 'Segoe UI', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
+@ff-sans: -apple-system, BlinkMacSystemFont, 'Segoe UI', Helvetica, Cantarell, Ubuntu, Roboto, Arial, sans-serif,
+          'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
```
francisrupert, line 70: Now that I consider the order, I'm assuming the inadvertent character rendering those screenshots was because it wound up using one of the Emoji fonts for everything. And thus we need to declare system fonts for as many platforms as possible?

josedialpad (approved):

lib/build/less/variables/typography.less
braddialpad: Pretty much yeah. On ubervoice they actually use the Joypixels font stack so they manually insert it after "sans-serif" in our stack. Because Joypixels is not a system font and is included in our web bundle it "always" downloads and displays that one if there are no other fonts earlier in the stack available.

At some point we'll be rid of the joypixels font, and use dialtone emojis instead which use images. Unfortunately a large migration would have to be performed in the product to do this.

braddialpad:
Changing branch to staging now that version7 has been merged.

PR #716, opened 2022-10-21
feat: add presence

modialpad:
## Description

Add `d-presence` styles indicating user statuses. 

A`<Presence/>` component in `dialtone-vue` will directly use these styles. 

## Pull Request Checklist

GI

### DIFF
--- a/docs/components/presence.md
+++ b/docs/components/presence.md
@@ -2,7 +2,7 @@
 title: Presence
 desc: A visual control element indicating the current status of a user
 ---
-<code-well-header>
+<code-well-header bgclass='d-bgc-white'>
   <div class="d-d-flex d-ai-center">
     <example-presence presence="active"/>
   </div>
--- a/docs/components/presence.md
+++ b/docs/components/presence.md
@@ -41,9 +41,7 @@ desc: A visual control element indicating the current status of a user
 
 ```html
 <div class="d-presence">
-  <div
-    class="d-presence__inner d-presence__inner--busy"
-  />
+  <div class="d-presence__inner d-presence__inner--busy" />
 </div>
 ```
 
@@ -57,9 +55,7 @@ desc: A visual control element indicating the current status of a user
 
 ```html
 <div class="d-presence">
-  <div
-    class="d-presence__inner d-presence__inner--away"
-  />
+  <div class="d-presence__inner d-presence__inner--away" />
 </div>
 ```
 
@@ -73,9 +69,7 @@ desc: A visual control element indicating the current status of a user
 
 ```html
 <div class="d-presence">
-  <div
-    class="d-presence__inner d-presence__inner--offline"
-  />
+  <div class="d-presence__inner d-presence__inner--offline" />
 </div>
 ```
 
--- a/docs/components/presence.md
+++ b/docs/components/presence.md
@@ -1,0 +1,82 @@
+---
+title: Presence
+desc: Presence is a visual control element indicating the active state of the user
+---
+<code-well-header>
+  <div class="d-d-flex d-ai-center">
+    <example-presence presence="active"/>
+  </div>
+</code-well-header>
+
+## Usage
+
+The Presence component renders a circle indicating showing the status of a user.
+
+<dialtone-usage>
+<template #do>
+
+- When its action has an instantaneous effect.
+</template>
+<template #dont>
+
+- When its action does not have an immediate effect on the application.
+- Selecting between 2 options. Instead, utilize a [Checkbox](checkbox.md).
+- As an alternative to a [Checkbox](checkbox.md) or [Radio](radio.md) within a Form.
+</template>
+</dialtone-usage>
+
+### Best practices
+
+- A Presence component should be used as a way to show the current status of a user.
+- A `presence` prop can be used to determine the status of the user, with the following valid statuses: `away`, `busy`, `active`, `offline`
+
+## Variants and examples
+
+### Base Styles
+
+<code-well-header>
+  <fieldset class="d-stack8">
+    <div class="d-d-flex d-ai-center">
+      <example-presence presence="active"/>
+    </div>
+    <div class="d-toggle-group d-d-flex d-ai-center">
+      <example-presence presence="busy"/>
+    </div>
+    <div class="d-toggle-group d-d-flex d-ai-center">
+      <example-presence presence="away"/>
+    </div>
+    <div class="d-toggle-group d-d-flex d-ai-center">
+      <example-presence presence="active"/>
+    </div>
+  </fieldset>
+</code-well-header>
+
+```html
+<fieldset class="d-stack8">
+  <div class="d-d-flex d-ai-center">
+    <example-presence presence="active"/>
+  </div>
+  <div class="d-toggle-group d-d-flex d-ai-center">
+    <example-presence presence="busy"/>
+  </div>
+  <div class="d-toggle-group d-d-flex d-ai-center">
+    <example-presence presence="away"/>
+  </div>
+  <div class="d-toggle-group d-d-flex d-ai-center">
+    <example-presence presence="active"/>
+  </div>
+</fieldset>
+```
+
+## Classes
+
+<component-class-table component-name="presence" />
+
+## Accessibility
+
+The best accessibility is semantic HTML. Most screen readers understand how to parse buttons if they’re correctly formatted.
+
+<script setup>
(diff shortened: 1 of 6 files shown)


docs/components/presence.md
```diff
@@ -0,0 +1,87 @@
...
+</div>
+```
+
+## Classes
+
+<component-class-table component-name="presence" />
+
+## Accessibility
+
+The best accessibility is semantic HTML. Most screen readers understand how to parse buttons if they’re correctly formatted.
```
braddialpad: > Most screen readers understand how to parse buttons if they’re correctly formatted.

Is this relevant to this component since it's not a button?

Since the component is just a visual indicator do we need to provide instruction to the developer using it on how it should be read out via screen reader? (aria-label if interactive, maybe sr-only class if not)

lib/build/less/components/presence.less
```diff
@@ -0,0 +1,47 @@
...
+//  visit https://dialpad.design/components/toggle
+
+.d-presence {
+  position: relative;
+  display: inline-block;
+  box-sizing: border-box;
+  border-color: var(--black-100);
+  border-style: solid !important;
+  border-width: var(--su2) !important;
+  border-radius: 50% !important;
```
braddialpad: hmm why `!important`? We only usually set this on utility classes that are meant to be immutable.

docs/components/presence.md
```diff
@@ -0,0 +1,87 @@
...
+title: Presence
+desc: A visual control element indicating the current status of a user
+---
+<code-well-header bgclass='d-bgc-white'>
+  <div class="d-d-flex d-ai-center">
+    <example-presence presence="active"/>
+  </div>
+</code-well-header>
+
+### Best practices
```
braddialpad, line 13: Best practices is part of the `## Usage` section

docs/components/presence.md
```diff
@@ -0,0 +1,87 @@
...
+
+- A Presence component should be used as a way to show the current status of a user.
+- A `presence` prop can be used to determine the status of the user, with the following valid statuses: `away`, `busy`, `active`, `offline`
```
braddialpad: "a css class can be set", remember as of right now dialtone knows nothing about vue.

docs/components/presence.md
```diff
@@ -0,0 +1,87 @@
...
+---
+<code-well-header bgclass='d-bgc-white'>
+  <div class="d-d-flex d-ai-center">
+    <example-presence presence="active"/>
+  </div>
+</code-well-header>
+
+### Best practices
+
+- A Presence component should be used as a way to show the current status of a user.
```
braddialpad: Is this all it should be used for? Could this component be used to show status for something that is not a user?

lib/build/less/components/presence.less
```diff
@@ -0,0 +1,47 @@
...
+
+//
+//  DIALTONE
+//  COMPONENTS: PRESENCE
+//
+//  These are presence classes for Dialpad's design system Dialtone.
+//  For further documentation of these and other classes,
+//  visit https://dialpad.design/components/toggle
+
+.d-presence {
```
braddialpad, line 11: Is there a figma for this? Just want to verify the css.

braddialpad (changes requested):
Thanks for this Mo!

lib/build/less/components/presence.less
modialpad: I don't think there is an independent Figma just for this. I got the general idea of this design from the [Leftbar Figma](https://www.figma.com/file/6n4RCYk2y2rYx5c4hL088f/DP-2.0-Left-sidebar?node-id=1%3A13198) and then confirmed with @francisrupert through this [Jira ticket](https://dialpad.atlassian.net/browse/DT-745) the requirements of this component.

docs/components/presence.md
modialpad: Good question, AFAIK the version of presence developed here this is meant to eventually be embedded as part of Avatar. I don't think it will be useful anywhere else but @francisrupert can confirm. It's possible that in the future we can extend the styles for presence for other use cases.

docs/components/presence.md
modialpad: Must have missed this while copying off of another template.

The only accessibility feature we'll need that this relevant to this is the announcement of any changes to the user's presence. It's not going to be interactive, so I think `sr-only` makes sense.

In the Vue component, I'm thinking we would add `role=status` [attribute](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA22) and then append a `<span>` that has the sr-only text, like so:

```js
<div v-if="presence" class="d-presence" role="status">
 <span class="sr-only">{{ srText }} </span>
  <div
    class="d-presence__inner"
    :class="{
      'd-presence__inner--active': presence === 'active',
      'd-presence__inner--away': presence === 'away',
      'd-presence__inner--busy': presence === 'busy',
      'd-presence__inner--offline': presence === 'offline',
    }"
  />
</div>
```

lib/build/less/components/presence.less
modialpad: Removed

docs/components/presence.md
modialpad: Done

docs/components/presence.md
modialpad: Re-wrote this to mention CSS classes

docs/components/presence.md
braddialpad: Sounds good. Might want to have a way to make this optional since the developer might not always want presence to be aria-live announced. Suppose it could just be if the srText was falsey.

docs/components/presence.md
braddialpad: Since this is such a standalone indicator not really tied to anything I could see it used for other things in the future. This is fine for now though.

docs/components/presence.md
braddialpad: May want to just write in this section "You may wish to announce any live changes to this component via screen reader since this is only a visual indicator".

lib/build/less/components/presence.less
braddialpad: Thanks that's all I was looking for really.

Are you going to implement that "transparent border" around the presence indicator as part of the avatar?

braddialpad (approved):
I had a couple final comments but otherwise this is good to go.

lib/build/less/components/presence.less
modialpad: Isn't it already part of `d-presence` styles? `d-presence` creates the a greyish border and `d-presence__inner` will be the circle indicator itself.

For the Avatar work next, I'll basically be shoving a `<Presence/>` component inside Avatar and expose it through a prop. I don't have a Jira defined for this yet but I will create it soon.

modialpad:
@braddialpad when I commit locally, the `lint:code` command produces a few thousand errors:
image

But on CI it's only a few warnings... I did notice a commit from @juliodialpad downgrading eslint, but I have version 8.22.0 and I still see this:

image

juliodialpad:
@modialpad what version of `node` are you using? I think we are using `node  v16.x` that might have to do with the error.

modialpad:
> @modialpad what version of `node` are you using? I think we are using `node v16.x` that might have to do with the error.

@juliodialpad I was using v16.14.0. I tested it out with CI's node version (v16.17.1 & npm 8.15.0) and I still see the same issue locally.

juliodialpad:
> > @modialpad what version of `node` are you using? I think we are using `node v16.x` that might have to do with the error.
> 
> @juliodialpad I was using v16.14.0. I tested it out with CI's node version (v16.17.1 & npm 8.15.0) and I still see the same issue locally.

I'm using v16.14.2, pulled your branch, did `npm install` and then `npm run lint:code`

image

And just having 2 warnings... did you removed `package-lock.json` or something?

braddialpad:
I'd try all the usual culprits, delete node_modules, delete package-lock, then npm install again

modialpad:
> I'm using v16.14.2, pulled your branch, did `npm install` and then `npm run lint:code`
> 
> image
> 
> And just having 2 warnings... did you removed `package-lock.json` or something?

Ok, may have found the cause. Apparently all the errors were happening under `docs/_site`. That's a generated folder, I had created it previously on the repo using the command `build:docsite`.  

The problem is that `eslintignore` did not include the folder `_site` as well, we only added it to .gitignore. So when I ran the lint command it also linted that whole folder.

I think the `_site` folder is not supposed to be available locally anyway but in any case `_site` should be definitely added to `eslintignore`.

PR #718, opened 2022-10-26
feat: publish v7 general icons

juliodialpad:
## Description

Removed Unconverted Icons and replaced with @francisrupert Manually converted ones.
Added build step to copy this converted icons to dist.

## Pull Request Checklist


gulpfile.js
```diff
@@ -601,6 +607,7 @@ const transformStrokeToFill = function (done) {
           .transform(
               `${paths.svgs.newInputRoot}/${category}/`,
               `${paths.svgs.newOutputRoot}/${category}/`,
+              { traceResolution: 600, showProgressBar: true },
```
braddialpad, line 628: Do we need this anymore? We aren't transforming anymore right, should we just remove the library and all the code related to it or am I not understanding the process correctly?

braddialpad:
How would you actually import these tokens if you were using the package, would it be a subdir?

gulpfile.js
```diff
@@ -637,11 +644,10 @@ const buildNewSVGIcons = function (done) {
   if (!settings.svgs) return done();
 
   //  Compile icons
-  return src(`${paths.svgs.newInputRoot}/**/*.svg`)
-  // replace any instances of the primary color in SVG with the theme class
+  return src(paths.svgs.version7Input)
     .pipe(addAttrsToSVG())
```
francisrupert: @juliodialpad,

Each of the new `dist` icons have `class="ICON_NAME"`, which is risky, e.g. `class="tag"`. 

Since we're going to use `d-icon` and `d-icons--SIZE`, we may as well prepend `d-icon` on the `<svg>` file itself. Within `addAttrsToSVG()` do `class="d-icons--${converted}".

(although, that might be more relevant to do in DT-721)

gulpfile.js
juliodialpad: We are going to use it to transform icons from `/newIcons` gitignored directory and save it to `/lib/build/svg/v7` I'll add the documentation

juliodialpad:
> How would you actually import these tokens if you were using the package, would it be a subdir?

Yes, right now is a subdir `@dialpad/dialtone/lib/dist/svg/v7/**/*.svg` That approach is ok or would it be better to export the `Vue` version of the icons?

gulpfile.js
juliodialpad: Yes, I'm adding that into https://github.com/dialpad/dialtone/pull/719

braddialpad:
> > How would you actually import these tokens if you were using the package, would it be a subdir?
> 
> Yes, right now is a subdir `@dialpad/dialtone/lib/dist/svg/v7/**/*.svg` That approach is ok or would it be better to export the `Vue` version of the icons?

hmmm we should probably do vue versions since that's how developers currently use the icons in product. Might be a bit confusing if they have to switch to SVG.

gulpfile.js
braddialpad: You mentioned in this PR that Francis had already manually converted them via figma, is there still further conversion needed via this transform?

juliodialpad:
> > > How would you actually import these tokens if you were using the package, would it be a subdir?
> > 
> > 
> > Yes, right now is a subdir `@dialpad/dialtone/lib/dist/svg/v7/**/*.svg` That approach is ok or would it be better to export the `Vue` version of the icons?
> 
> hmmm we should probably do vue versions since that's how developers currently use the icons in product. Might be a bit confusing if they have to switch to SVG.

Vue implementation is added on https://github.com/dialpad/dialtone/pull/719 which has this branch as its base.

braddialpad:
Great thanks!

francisrupert (commented):
As far this specific PR goes, I have nothing further, especially with #719 being the real result. Remaining discussion appears to be Vue-centric, so I'll defer to @braddialpad for PR approval.

gulpfile.js
juliodialpad: Yes, we were discussing with @francisrupert that there might be some icons we would like to convert using this technique as Figma has some bugs too.

braddialpad (approved):

PR #721, opened 2022-10-28
fix: correct "busy" color

francisrupert:
## Description
* Correct red "busy" color of Presence from `400` to `300`
* Made use of local css custom properties

/c @modialpad 

## Pull Request Checklist

image

### DIFF
--- a/lib/build/less/components/presence.less
+++ b/lib/build/less/components/presence.less
@@ -9,38 +9,45 @@
 //  visit https://dialpad.design/components/presence
 
 .d-presence {
-  position: relative;
+  --color-border-base: var(--black-100);
+  --color-border-offline: var(--black-400);
+  --color-background-base: var(--white);
+  --color-background-active: var(--green-400);
+  --color-background-busy: var(--red-300);
+  --color-background-away: var(--gold-200);
+  --color-background-offline: var(--white);
+
   display: inline-block;
   box-sizing: border-box;
-  border-color: var(--black-100);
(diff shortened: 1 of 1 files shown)


braddialpad (approved):

lib/build/less/components/presence.less
```diff
@@ -9,38 +9,45 @@
...
 
 .d-presence {
-  position: relative;
+  --color-border-base: var(--black-100);
+  --color-border-offline: var(--black-400);
+  --color-background-base: var(--white);
+  --color-background-active: var(--green-400);
+  --color-background-busy: var(--red-300);
+  --color-background-away: var(--gold-200);
+  --color-background-offline: var(--white);
```
modialpad, line 18: LGTM, I wasn't sure if we're doing this here or in DT-Vue but I think this is better 👍 

One question is about naming, do these `--color-background-*` need to be prefixed with something that indicate that they're owned by presence? `--color-background-offline` can sound a bit generic

francisrupert: @modialpad,

It's scoped, and can't be used outside `.d-presence {}`. As a tiny component with simple CSS I'm ok with it. We don't currently have strict naming rules for _local_ CSS Custom Properties. For more elaborate locally-scoped CSS Custom Properties we would be more specific with the. name e.g. `--button-color-background-secondary: pink`.

Once Design Tokens are formally put into play we'll probably have a prescribed naming convention for local CSS Custom Properties.

PR #737, opened 2022-11-17
feat: what's new blog page

braddialpad:
## Description

Finally managed to get this PR up.

Kept this very basic design-wise since we didn't make designs for this, and I'm sure you'll have much better opinions than me @francisrupert.

Tried to keep the editing as pure markdown as possible, you do currently have to wrap your main content in a vue component in the markdown file, but it will be the same structure on every page so you just have to copy paste it for the time being. Heading / author / date is all set via frontmatter. To add a new blog post, simply create a new markdown file in the `about/whats_new/posts` folder with the current date in `yyyy-m-d` format. This markdown will be dynamically added to the index page as well as it's own dedicated page with it's own route.

There is one fake test blog post in there and one real one. I will remove the fake before merging.

Also my first attempt at using the composition API, I think it went okay. Had to add a bit of hacky css in here because our overrides for h2 / h3 etc were not working with the HTML structure I added, but it should be isolated to this component. Had some technical issues with getting the first paragraph out of the main content and into the preview. The preview just shows the heading for now.

This is pretty barebones, lots of room for iterate in the future.

- Getting rid of the vue component in the markdown file so it's just pure markdown, or just making a UI to create blog posts.
- Improved visual design
- Pagination for historical posts
- Could add more buttons for sharing etc, just have copy link for now
- Add initial paragraph or a description to preview

new libraries:
- `date-fns` because vanilla javascript REALLY sucks at dates
- ~`vite-plugin-md` allows importing of `.md` files~ replaced with Julio's clever solution

## Pull Request Checklist


josedialpad:
Changes look good overall. Wondering if there's any config in the [`vite-plugin-md` plugin](https://github.com/antfu/vite-plugin-md) we need to use to fix the issue in the build? Or maybe in the vitepress build config?
Can't understand the error well, an idea is that during the conversion of the `md` into the vue component, the router link is not provided?

braddialpad:
Still failing even after Julio's fix.. super weird.

braddialpad:
I'm pretty convinced this is an SSR related issue since the error is now `ReferenceError: document is not defined`.

I tried wrapping some things in `<ClientOnly>` but hasn't helped so far.

braddialpad:
Issues with the build all fixed! Ready for review.

docs/.vuepress/baseComponents/BlogPostPreview.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <dt-link
+    class="d-td-none d-fc-unset d-d-block"
+    :href="$withBase(`/about/whats_new/posts/${format(posted, 'y-M-d')}`)"
+  >
```
juliodialpad: Is there a way to change this to a `<router-link>` ? I don't like having internal links as externals because we lose animations and it needs to reload all the site.

juliodialpad: Something like: 
```suggestion
  <router-link
    class="d-link d-td-none d-fc-unset d-d-block"
    :to="$withBase(`/about/whats_new/posts/${format(posted, 'y-M-d')}`)"
  >
```

juliodialpad (changes requested):
Looks really good, thanks @braddialpad!

Just check if you can replace that link with a router-link

docs/.vuepress/baseComponents/BlogPostPreview.vue
braddialpad: I did it with a custom router-link so we could still use dt-link, seems to work nicely

braddialpad: also prevents having to use $withBase

docs/.vuepress/theme/components/Home.vue
```diff
@@ -55,8 +55,8 @@
               size="lg"
               importance="outlined"
               class="d-ml16"
-              @click="(e) => { sendAnalyticsEvent(); navigate(e); }"
-              @keypress.enter="(e) => { sendAnalyticsEvent(); navigate(e); }"
+              @click="(e) => { navigate(e); }"
+              @keypress.enter="(e) => { navigate(e); }"
```
juliodialpad, line 59: Aren't we using Google Analytics any more?

docs/.vuepress/baseComponents/BlogPostPreview.vue
```diff
@@ -1,21 +1,28 @@
...
+    v-slot="{ navigate }"
+    :to="`/about/whats_new/posts/${format(posted, 'y-M-d')}`"
+    custom
   >
-    <dt-card
-      class="d-mt72 h:d-bgc-black-200"
+    <dt-link
+      class="d-td-none d-fc-unset d-d-block"
+      href="`/about/whats_new/posts/${format(posted, 'y-M-d')}`"
+      @click="(e) => { navigate(e); }"
```
juliodialpad: Do we need to have `:to` and `href` here?

juliodialpad (commented):
it works like a charm! just 2 comments

docs/.vuepress/baseComponents/BlogPostPreview.vue
braddialpad: I think we actually do because it's a custom element, the to is what actually navigates to the link using the router, but if I just remove the href from dt-link:

- it is no longer focusable by keyboard
- the `<a>` tag in the dom shows no href at all, which isn't really correct
- all the other `<a>` tags on the site have a href with the link even if they are using router-link. router-link will do this automatically if you're not using a custom

docs/.vuepress/theme/components/Home.vue
braddialpad: I don't know it only seems to be added on the one link on the homepage, I think that was the POC don't think we setup anything more than that

docs/.vuepress/baseComponents/BlogPostPreview.vue
juliodialpad: Sounds good, lets keep it then, thanks for the explanation!

juliodialpad (approved):

docs/.vuepress/theme/components/Home.vue
juliodialpad: Right, I think that event was a POC to check that events could be logged in Analytics. So it's ok to remove it.

We are still getting a lot of insight from it, we just need to keep checking it to improve our site.

PR #752, opened 2022-11-29
docs: fix async loading for dialtone vue

braddialpad:
## Description
Add the same async fix @juliodialpad put into the new theme. Just putting it here to release earlier.


juliodialpad (approved):
Looks good! that will fix the issues, thanks Brad

PR #770, opened 2022-12-14
chore: add css/less sourcemaps to local vite

braddialpad:
## Description
Add css / less sourcemaps so we can see which file a style was set from.

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -41,6 +41,9 @@ export default defineUserConfig({
 
   bundler: viteBundler({
     viteOptions: {
+      css: {
+        devSourcemap: true,
+      },
       resolve: {
         alias: {
           vue$: 'vue/dist/vue.esm-bundler.js',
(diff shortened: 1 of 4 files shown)


francisrupert (commented):
This is awesome. To be honest I've implement them for Sass and Stylus in Grunt and Gulp... but only just "mostly" — it's never 100% perfect. 

I tested it and it does exactly what I want, but sometimes clicking on the source urls 404s. See screenshot for examples. This is the part that always got me and I gave up on it and just accepted it. Not the worst thing to have missing. If it's not solvable reasonably quickly, we can probably use as is and log low-priority task to resolve.

source-map

braddialpad:
> This is awesome. To be honest I've implement them for Sass and Stylus in Grunt and Gulp... but only just "mostly" — it's never 100% perfect.
> 
> I tested it and it does exactly what I want, but sometimes clicking on the source urls 404s. See screenshot for examples. This is the part that always got me and I gave up on it and just accepted it. Not the worst thing to have missing. If it's not solvable reasonably quickly, we can probably use as is and log low-priority task to resolve.

Nice find, yeah the paths were completely wrong. managed to fix this with the `sourcemaps.mapSources()` function.

francisrupert (approved):
Yep, works a charm! Thx!

PR #782, opened 2023-01-07
docs: add component change history in about page

josedialpad:
## Description
- Parsing and showing releases (this PR).
- Generation of the changelog.json file as part of the release pipeline is in progress https://github.com/dialpad/dialtone/pull/794.

## Pull Request Checklist


package.json
```diff
@@ -126,6 +127,7 @@
     "lint-staged": "^13.1.0",
     "markdownlint": "^0.26.2",
     "markdownlint-cli": "^0.32.2",
+    "marked": "^4.2.12",
```
braddialpad: vuepress includes markdown-it which does this, so we don't need to include another library.

We had a component to render markdown before: https://github.com/dialpad/dialtone/commit/0fc9bd0a6d6805c913831314b835b823703c6505

But we removed it because we thought it was bad practice to render markdown within vue files. Though in this case I'm not sure there's any way around it, is there @juliodialpad?

docs/_data/changelog.json
```diff
@@ -0,0 +1,1836 @@
+{
```
braddialpad: Is there supposed to be 2 `changelog.json`?

braddialpad (changes requested):

docs/_data/changelog.json
josedialpad: no, the only needed is in the project root. Removed this one as it was just used for testing purposes.

package.json
juliodialpad: Yes, including two renderers is a bad practice.

Maybe you can write this in a .md file https://v2.vuepress.vuejs.org/advanced/cookbook/markdown-and-vue-sfc.html

And that way you just need to write {{ formatReleaseNotes(item) }} and Vuepress will render the markdown there instead of having to add another renderer.

josedialpad: good call, replaced the marked library with markdown-it. Seems it's a workaround used by the community https://github.com/vuejs/vuepress/issues/257#issuecomment-897111820.

The [workaround using slots](https://github.com/vuejs/vuepress/issues/257#issuecomment-384336255) proposed by the maintainer didn't work in this case where you're not passing the raw markdown text in the slot but rather passing from a prop.

> Maybe you can write this in a .md file https://v2.vuepress.vuejs.org/advanced/cookbook/markdown-and-vue-sfc.html

That works but faced two issues: 1. importing this md file in the dialtone.md page (maybe we can import with [this plugin](https://plugin-md-enhance.vuejs.press/guide/include.html)?) and 2. the usage of the props with this approach.
Putting all the code in the dialtone.md will work but will be a mess ([example](https://github.com/dialpad/dialtone/pull/782/commits/9e0bbe62a2ab405bdec6bcd1076d55c22e2337c0)). Any ideas? 

Wondering since Vuepress uses the same `markdown-it` library to compile the markdown into HTML (and we're doing it manually in the MarkdownRender.vue component) and then converts to a vue SFC which we already have with the `DialtoneChangelog.vue` isn't it the same process?

braddialpad: > importing this md file in the dialtone.md page (maybe we can import with [this plugin](https://plugin-md-enhance.vuejs.press/guide/include.html)?)

Yeah I have had struggles with this too. the plugin might be a good idea if it works.

> Wondering since Vuepress uses the same markdown-it library to compile the markdown into HTML (and we're doing it manually in the MarkdownRender.vue component) and then converts to a vue SFC which we already have with the DialtoneChangelog.vue isn't it the same process?

I think Julio is saying you don't have to use a MarkdownRender.vue component at all if you do it as he mentioned.

josedialpad: > Yeah I have had struggles with this too. the plugin might be a good idea if it works.

It worked and added a [POC commit](https://github.com/dialpad/dialtone/commit/8498ffaf89f2b305be33689b6b0ae34724ff0878).

> I think Julio is saying you don't have to use a MarkdownRender.vue component at all if you do it as he mentioned.

Yeah that makes sense. Unfortunately it doesn't look like a good approach to use here or in complex cases. The usage of the props with this approach is still a problem as it's required to show proper release notes based on the project prop passed in the component. Otherwise we need to duplicate the md files with the release notes of each project (dialtone, dialtone-vue) which will violete the DRY principle.

braddialpad: Right I think we can just go with the markdown renderer method then, just add a note not to use it except in very specific cases

josedialpad: added doc https://github.com/dialpad/dialtone/commit/b2ae3fc70b13ee044b9d409a94ac54e0da867cb1

docs/.vuepress/views/DialtoneChangelog.vue
```diff
@@ -0,0 +1,72 @@
...
+  project: {
+    type: String,
+    default: 'Dialtone',
+    validator (value) {
+      return ['Dialtone', 'DialtoneVue'].includes(value);
+    },
+  },
+});
+
+const changelogJson = computed(() => props.project === 'DialtoneVue' ? dialtoneVueChangelog : dialtoneChangelog);
```
braddialpad: Might as well create a mapping for this. It will scale much better into the future for example if we need to add other repos (dialtone-icons, dialtone-tokens for example). You can use this mapping object for the prop validator and this change log mapping
```
const mapping = {
  DialtoneVue: dialtoneVueChangelog,
  Dialtone: dialtoneChangelog,
}
```

```
const changelogJson = computed(() => mapping[props.project]);
```

braddialpad: Or maybe even a [Map object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) if you prefer

josedialpad: Good call, added a constant variable https://github.com/dialpad/dialtone/pull/782/commits/4fadbab2302c3fa3ad59b5081ea8ef83597c4760

docs/.vuepress/views/DialtoneChangelog.vue
```diff
@@ -0,0 +1,71 @@
...
+});
+
+const changelogJson = computed(() => DIALTONE_CHANGELOGS[props.project]);
+
+const getVersion = (item) => changelogJson.value.versions[item].version;
+
+const getGithubReleaseUrl = (item) => `https://github.com/dialpad/dialtone/releases/tag/v${getVersion(item)}`;
+
+const formatReleaseNotes = (note) => {
+  const noteWithoutExtraAsterisks = note.replace(/\*\*/g, '');
```
braddialpad: Maybe put this in a function as well, I know it's simple but just to be consistent with the other ones.

docs/.vuepress/views/DialtoneChangelog.vue
```diff
@@ -0,0 +1,71 @@
...
+  const releaseNoteWithCommitLink = formatReleaseNotesWithCommitLink(noteWithoutExtraAsterisks);
+  const releaseNoteWithPrLink = formatReleaseNotesWithPrLink(releaseNoteWithCommitLink);
+  return releaseNoteWithPrLink;
```
braddialpad: Optional minor improvement, if you happen to prefer this. Isolate the note value and the functions within a object, and then chain them.

```js
var ReleaseNoteFormatter = {
  note: 'change log here',
  withoutAsterisks: function() {
    return this.note.replace(/\*\*/g, '');
  },

  withCommitLink: function() {
    ...
  },
 
  withPrLink: function() {
    ...
  }
}
```

```js
ReleaseNoteFormatter.note = note;
ReleaseNoteFormatter.withoutAsterisks().withCommitLink().withPrLink();
```

docs/about/dialtone.md
```diff
@@ -3,6 +3,37 @@ title: About Dialtone
...
+    >
+      <img
+        alt="Dialtone Vue version number"
+        src="https://img.shields.io/github/package-json/v/dialpad/dialtone-vue/vue3?color=D3BCFF&label=Vue"
+      >
+    </a>
+    <dialtone-changelog project="DialtoneVue" />
+  </div>
+</div>
+
```
braddialpad, line 36: Shouldn't we write this in markdown or is there something we can't achieve that way?

docs/.vuepress/views/DialtoneChangelog.vue
braddialpad: or even add another function within there ReleaseNoteFormatter.format() that just calls all 3

docs/.vuepress/views/DialtoneChangelog.vue
josedialpad: updated https://github.com/dialpad/dialtone/pull/782/commits/04c6561bd0c4651ef1f025dc2b90df2ea217c7e8

docs/.vuepress/views/DialtoneChangelog.vue
josedialpad: good call, updated to object syntax and pulled into the utilities https://github.com/dialpad/dialtone/pull/782/commits/249c17629c05161ec9609dfae11f1489751d9162

docs/about/dialtone.md
josedialpad: Yeah ideally write in markdown but the table can be tricky as you cannot style it too much AFAIK. Also you can't mix or embed markdown inside the HTML so that's I went with full html here

braddialpad (approved):
Looks much cleaner, thanks!

PR #785, opened 2023-01-09
chore: staging preview

braddialpad:
Deploy a staging preview via GHA

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -6,6 +6,7 @@ on:
       - production
       - alpha
       - beta
+      - staging
 
 jobs:
   deploy:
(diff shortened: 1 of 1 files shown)


francisrupert (approved):
lgtm

.github/workflows/deploy.yml
```diff
@@ -36,11 +37,17 @@ jobs:
...
-        if: ${{ github.ref == 'refs/heads/production' }}
         run: npm run build:docsite
 
+      - name: Deploy to GitHub Pages (Staging)
+        if: ${{ github.ref == 'refs/heads/staging' }}
+        uses: JamesIves/github-pages-deploy-action@v4.2.5
+        with:
+          branch: gh-pages
+          folder: docs/.vuepress/dist
+          target-folder: staging
```
josedialpad, line 49: Shouldn't there be a `clean-exclude` for other folders here, like:

```
          clean-exclude: |
            deploy-previews
            version6
            production
```

josedialpad (approved):
w/ one question. :shipit:

.github/workflows/deploy.yml
braddialpad: not in this case because of `target-folder: staging`. It only auto cleans folders WITHIN the specified target folder.

PR #809, opened 2023-02-02
docs: components health status page

juliodialpad:
## Description

Created a POC with components health status page, added documentation about the frontmatter properties we are using.

## Pull Request Checklist


docs/.vuepress/theme/index.js
```diff
@@ -50,6 +50,33 @@ function extractFrontmatter (app, path, options) {
     .sort((a, b) => sortingArr.indexOf(a.link) - sortingArr.indexOf(b.link));
```
iropolo, line 55: I suggest to split these code and create sth like

const pathPages = getPagePath(whatever, 'path');
const pathPagesFiltered = getHtmlPages(pathPages);
const htmlFrontmatterPages = getFrontmatter(pathPagesFiltered);
const htmlFrontmatterPagesFormated = set....
const sorted..... = sort....

Create a folder called utils and move those function there.

It will be more readable and scalable.
Apply this for the following code.

If `blogPostsFrontmatter` , `extractFrontmatter` , `extractComponentStatus` are local functions and they will not be exported, I would refactor those with _ on the beggining or move them to util foolder.

This are just suggestions and it is not necessary to be applied at this moment. :)

iropolo:
I really like this POC :)

docs/.vuepress/theme/index.js
```diff
@@ -50,6 +50,33 @@ function extractFrontmatter (app, path, options) {
...
 }
 
+function extractComponentStatus (app) {
+  const indexPage = app.pages.find(page => page.path === '/components/status/');
```
braddialpad, line 59: While we're at it we might as well create new pages for our two planned components this quarter "Emoji Picker" and "Infinite Scroll".

.github/CONTRIBUTING.md
```diff
@@ -287,6 +289,26 @@ If you are developer contributing to Dialtone but are not on the Dialtone team y
...
+We use markdown frontmatter to add metadata to the pages, and we have several functions to extract such data and manipulate it.
+Here's an overview of important properties and the values they need/can have:
+
+- title `(required)`: Used as the page title, and also as the component name for [Components overview page].
+- shortTitle `(optional)`: This property is used to fix linking issues on [Components overview page] when the title is different from the component name.
+- desc `(optional)`: Used as the page subtitle.
+- status `(optional)`: CSS Component status, used to display a badge on [Components overview page] and
+also to define the component status on [Components status page], 
+  - Options available: `['wip', 'planned', 'new', null]` 
+  if status is not defined, the components will have a 'No implemented Yet' status.
```
braddialpad: "Not implemented yet"

docs/.vuepress/theme/index.js
```diff
@@ -50,6 +50,33 @@ function extractFrontmatter (app, path, options) {
...
+          default:
+            return 'DONE';
+        }
+      };
+      return {
+        url: page.path,
+        name: frontmatter.title,
+        figma: componentStatus(frontmatter.figma),
+        vue: componentStatus(frontmatter.storybook),
+        css: componentStatus(frontmatter.status),
```
braddialpad: I think for components such as "list-item" and "dropdown" that are vue only components and have no styles in dialtone we should set this to "not yet implemented".

braddialpad (commented):
Really awesome! I think we can probably just release this to production and then update when Francis is done with the final designs.

francisrupert (changes requested):
Yeah, this is awesome and I don't see much reason not to MVP release it. I'll eventually take some time for detailed feedback, but in the meantime:

* Page title: "Component status" (sentence case)
* I'm assuming the buttons are for filtering. I'll propose a UI for that later, so perhaps an MVP version of this doesn't have that functionality. In lieu of that, just list the icon/label pairing inline.
* The values you got from that old Confluence page is out-of-date. I started to effectively replace it with this [Spreadsheet](https://docs.google.com/spreadsheets/d/1mjQDvWAVAZV_qL6xmceu4WLeq7ZVimxWvzDDpUmooAE/edit#gid=0). 
  * Reconcile where you can can, and I'll validate through an updated PR. (ignore the columns for docs, a11y, iOS, Android.
* Reserve columns for Introduced/Modified version numbers, which we have separate task for.
* Change "Available" to "Ready"
* Change "Not implemented yet" with "Planned"
* Change "Work in progress" to "In progress"
* Add another value of "N/A". for example, a Figma file for "Lazy show" doesn't apply
* Make sure each table data cell has accessible equivalent of its value, e.g. "Ready / In progress / etc"

docs/.vuepress/theme/index.js
juliodialpad: Nice suggestion!

I'll add the underscore at the beginning cause this are local functions and we are not going to reuse it anywhere else, so I prefer to maintain them here.

Maybe a refactor could be done later as I notice I have some repeated code, but nothing too bad.

docs/.vuepress/theme/index.js
juliodialpad: Sure! Will do

docs/.vuepress/theme/index.js
juliodialpad: ~Sure, we only need to remove the `status` property from this components and will have that state or do you think it's a better pattern to set `status: NYI` to be more clear about the intention?~

When `status` property is not present in frontmatter, the component will have a `N/A`. Updated documentation with this changes too.

juliodialpad:
> * I'm assuming the buttons are for filtering. I'll propose a UI for that later, so perhaps an MVP version of this doesn't have that functionality. In lieu of that, just list the icon/label pairing inline.

Sure! Will do

> * The values you got from that old Confluence page is out-of-date. I started to effectively replace it with this [Spreadsheet](https://docs.google.com/spreadsheets/d/1mjQDvWAVAZV_qL6xmceu4WLeq7ZVimxWvzDDpUmooAE/edit#gid=0).
>   
>   * Reconcile where you can can, and I'll validate through an updated PR. (ignore the columns for docs, a11y, iOS, Android.

In fact, the values are being retrieved directly from the components frontmatter so they should be updated... I'll check the new list to see if there's any difference.

> * Reserve columns for Introduced/Modified version numbers, which we have separate task for.

Do you want me to add the columns now and leave them empty or just have in mind that we are adding new columns?


> * Add another value of "N/A". for example, a Figma file for "Lazy show" doesn't apply

Nice addition! Will add this.

> * Make sure each table data cell has accessible equivalent of its value, e.g. "Ready / In progress / etc"
Added `aria-label` and `tabindex=0` properties.

docs/.vuepress/theme/index.js
juliodialpad: Changed `list-item` and `dropdown` statuses to: planned.

juliodialpad:
@francisrupert The way I thought this is:

- If `figma` property is a URL: the component health status is "ready"... moving such component to `in progress` will remove the Figma button from the component page.

The same logic applies to `storybook` property.

I made it this way because it's simpler to have one property and make validations instead of having a lot of properties, e.g:

```md
---
status: ready
storybook: https://vue.dialpad.design/...
figma: https://www.figma.com/file/...
---
```

vs

```md
---
status: ready
storybook: ready
storybook_url: status: ready
figma: in progress
figma_url: https://www.figma.com/file/...
---
```

iropolo:
LGTM! 🚀

francisrupert (changes requested):
Here are updated values for Figma column. 

image

juliodialpad:
> Here are updated values for Figma column.
> 
> image

Can you please provide the link to `validation messages` Figma? I couldn't find it.

docs/components/validation-messages.md
```diff
@@ -3,7 +3,7 @@ title: Validation messages
 status: ready
 thumb: true
 desc: Validation messages are used to convey information to the user about the current state of the input element. These messages can have an error, warning or success type.
-storybook_url: https://vue.dialpad.design/?path=/story/components-validation-messages--default
+storybook: https://vue.dialpad.design/?path=/story/components-validation-messages--default
```
francisrupert, line 6: Figma link https://www.figma.com/file/2adf7JhZOncRyjYiy2joil/DT-Core%3A-Components-7?node-id=11399%3A76354&t=LqzEvQfr3DMHh7Og-11

francisrupert (approved):

PR #829, opened 2023-02-16
fix(popover): set default max height to never be larger than viewport

braddialpad:
## Description
Set the max height of the popover to never be larger than the viewport. Prior to this change data would be cut off without the ability to scroll.

Recommendation for this fix from popper.js: https://popper.js.org/docs/v2/faq/#my-popper-is-bigger-than-the-viewport-what-do-i-do

## Pull Request Checklist


braddialpad:
I think I may just put a intersect observer in the vue component. This is going to conflict too much with the maxHeight prop in dropdown.

braddialpad:
Closing

PR #850, opened 2023-03-16
chore: update to node 18

braddialpad:
Update project to use node 18


juliodialpad (approved):
If there are no build errors this is good to go, thanks @braddialpad

braddialpad:
> If there are no build errors this is good to go, thanks @braddialpad

Seems to be good, but I guess we won't truly know until we do a production deploy.

PR #855, opened 2023-03-24
fix(validation-messages): fix overlap with long message

braddialpad:
## Description

Had a team asking about this so decided to prioritize it since it's a pretty small fix
https://dialpad.atlassian.net/browse/DT-1024

Set min-width in addition to width on validation messages, otherwise the content expands to overlap the icon when there is a multiline message. This is because `background-image` is being used for the icon so it is not recognized as actual content within the element.

before:
Screenshot 2023-03-24 at 12 15 32 PM

after:
Screenshot 2023-03-24 at 12 15 38 PM

## Pull Request Checklist


lib/build/less/components/forms.less
```diff
@@ -114,7 +114,8 @@ fieldset {
   // Icon Slot
   &::before {
     display: block;
-    width: calc(var(--size-500) + var(--size-200)); // 16
```
braddialpad, line 117: @francisrupert this is 18 not 16 like the comment says. Is this intentional? the icon is 16x16.

francisrupert: Yeah, the `::before` is meant to be 16x16.

francisrupert:
> This is because background-image is being used for the icon...

And that's an outdated method of building this kind of thing. But swallowing that in the meantime since we're hoping all of these form-like components would be rebuilt (even though that also might be a while).

francisrupert (approved):

braddialpad:
> And that's an outdated method of building this kind of thing. But swallowing that in the meantime since we're hoping all of these form-like components would be rebuilt (even though that also might be a while).

Yep I agree. I was tempted to restructure this whole thing but held myself back for a quick fix for now.

PR #864, opened 2023-04-12
docs: dialtone tokens documentation

braddialpad:
## Description
First iteration of Dialtone tokens documentation. Very simple, only shows names and values. Examples will be much easier after we have done the migration to get Dialtone itself to use tokens so will defer that to then, and we don't have descriptions linked to tokens quite yet.

Included in this PR:

- viewing the different categories of tokens and their names and values
- changing the language / format of the tokens you are viewing which subsequently changes their names / values. Showing only css / android (kotlin) / iOS (swift) right now as they're the only ones currently important to us.
- clicking the copy button to copy the token name.

## Pull Request Checklist


docs/_data/site-nav.json
```diff
@@ -18,6 +18,29 @@
...
+          "text": "Sizing",
+          "link": "/tokens/sizing.html"
+        },
+        {
+          "text": "Spacing",
+          "link": "/tokens/spacing.html"
+        }
+      ]
+    }
+  ],
```
juliodialpad: We need to add "overview" section to the site-nav too as there's no way to go back to the Tokens overview rather than clicking "Tokens" on the top navigation.

image

docs/.vuepress/baseComponents/TokenTable.vue
```diff
@@ -0,0 +1,127 @@
...
+  </table>
+</template>
+
+<script>
+import CopyButton from './CopyButton.vue';
+
+const FORMAT_MAP = {
+  css: 'css/variables',
+  android: 'compose/object',
+  ios: 'ios-swift/enum.swift',
```
juliodialpad: If it's possible, change this to: 'CSS', 'Android', 'iOS'

juliodialpad (changes requested):
The way you automated this is so good! I love it but I don't love the `select-menu` I'd prefer to use tabs instead or a global selector if that's possible.

If we cannot use the tabs or a global selector, is it possible to fix that `select-menu` flickering?

https://user-images.githubusercontent.com/87546543/231551619-ce8f89d8-ca1e-41de-89e6-39df4f550fca.mov

docs/.vuepress/baseComponents/TokenTable.vue
```diff
@@ -0,0 +1,127 @@
...
+        if (key.split('/')[0] !== this.category) continue;
+        const token = {
+          name: value[FORMAT_MAP[this.format]]?.name,
+          description: '',
+          value: value[FORMAT_MAP[this.format]]?.value,
+        };
+        tokens.push(token);
+      }
+      return tokens;
+    },
```
iropolo, line 106: Code suggestion

```
tokensProcessed() {
  if (!this.json) return [];

  const tokens = Object.entries(this.json)
    .filter(([key]) => key.split('/')[0] === this.category)
    .map(([_, value]) => {
      const { name, value } = value[FORMAT_MAP[this.format]] || {};
      return { name, description: '', value };
    });

  return tokens;
}
```

iropolo: Where we are using this component?

iropolo: Code suggestion

```
tokensProcessed () {
      if (!this.json) return [];

      return Object.entries(this.json)
        .filter(([key]) => key.split('/')[0] === this.category)
        .map(([_, value]) => {
          const { name, value: tokenValue } = value[FORMAT_MAP[this.format]] || {};
          return { name, description: '', value: tokenValue };
        });
    },
```

docs/.vuepress/baseComponents/TokenTable.vue
```diff
@@ -0,0 +1,127 @@
...
+
+    selectMenuOptions () {
+      return Object.keys(FORMAT_MAP).map((item) => {
+        return { value: item, label: item };
+      });
+    },
+  },
+
+  beforeMount () {
+    import('../../../node_modules/@dialpad/dialtone-tokens/dist/doc.json').then((module) => {
```
iropolo: This works as well

`import('@dialpad/dialtone-tokens/dist/doc.json').then((module) => {`

docs/_data/site-nav.json
braddialpad: ahh yep missed that, thanks

docs/.vuepress/baseComponents/TokenTable.vue
braddialpad: yep much nicer thanks.

docs/.vuepress/baseComponents/TokenTable.vue
braddialpad: thanks for the improvment, works well!

docs/.vuepress/baseComponents/TokenTable.vue
braddialpad: done

docs/_data/site-nav.json
braddialpad: thanks, missed that.

braddialpad:
> but I don't love the `select-menu` I'd prefer to use tabs instead or a global selector if that's possible.

Yeah we didn't really have designs so thought i'd just throw something simple up there. I think some type of global selector would be ideal so it persists between pages, but I think that will be a future update. I'll at least change the width of this a bit so it's not so ridiculously huge.

> If we cannot use the tabs or a global selector, is it possible to fix that `select-menu` flickering?

Hmm weird. not sure why this is happening I can't really repro it, might be a problem with the component itself.

juliodialpad (approved):
Looking much nicer now, thanks! I can live with that little flickering select-menu 😂 no major issues found!

francisrupert:
Oh my goodness I'm finally seeing this. Frickin sweet. 

And yes, it's on me to dictate...
* where it lives (probably with [Design language](https://dialpad.design/design/) section)
* how the table should look
* what columns it has
* some good language introducing and surrounding it.

I'm also looking to update the `description` field currently mostly blank in [base.json](https://github.com/dialpad/dialtone-tokens/blob/staging/base.json#L135). That would populate like [it currently does ](https://user-images.githubusercontent.com/1165933/232154562-ae6f0494-51e5-46b6-9d79-8c205fa371ae.png) on https://dialpad.design/design/colors/#text.

In the meantime:

* It's not using the most current you cut a couple days ago: [1.7.0](https://github.com/dialpad/dialtone-tokens/releases/tag/v1.7.0).
* Let's suppress it from the global nav, but still allow it (secretly) be viewable.
* Provide a very basic swatch as first column, e.g. something like `<div class="d-bar-circle d-w42 d-h42" style="background-color:var(--foo);"></div>`.
  * I'm assuming it won't actually work since they're not actually imported, so at least a placeholder swatch I suppose.
* Can you format the `Name` column to what would actually be used in code, e.g. `var(--dt-color-surface-secondary)` for CSS, and whatever equivalent is for iOS/Android?

braddialpad:
> * Let's suppress it from the global nav, but still allow it (secretly) be viewable.

It's been removed from the nav but can still be accessed via dialpad.design/tokens



> Provide a very basic swatch as first column, e.g. something like `<div class="d-bar-circle d-w42 d-h42" style="background-color:var(--foo);"></div>`.
> * I'm assuming it won't actually work since they're not actually imported, so at least a placeholder swatch I suppose.

I've added a placeholder for now, can attempt applying tokens once they are imported and we start working on the migration next week.

> * Can you format the `Name` column to what would actually be used in code, e.g. `var(--dt-color-surface-secondary)` for CSS, and whatever equivalent is for iOS/Android?

done, no changes necesssary for iOS/Android but css has been fixed

> I'm also looking to update the `description` field currently mostly blank in [base.json](https://github.com/dialpad/dialtone-tokens/blob/staging/base.json#L135). That would populate like [it currently does ](https://user-images.githubusercontent.com/1165933/232154562-ae6f0494-51e5-46b6-9d79-8c205fa371ae.png) on https://dialpad.design/design/colors/#text.

Nice, didn't know we had this. It's now populated in the table and will update along with the tokens.

PR #867, opened 2023-04-19
feat(size-and-space): added half stops, negative equivalents, and applied to components

francisrupert:
## Description

* Half stops: e.g. `350` for "halfway between `300` and `400`". 
* Each size and space variable has a negative equivalent, e.g. `var(--space-350-negative)`
* Reduced the amount of `calc`'ing in favor of the half-stops. 

## Pull Request Checklist

image


docs/.vuepress/theme/assets/less/dialtone-docs.less
```diff
@@ -304,18 +304,18 @@ a.header-anchor {
   display: flex;
   flex-direction: column;
   justify-content: center;
-  width: calc((var(--icon-card-width) * 2) + var(--su32));
+  width: calc((var(--icon-card-width) * 2) + var(--size-600));
   min-height: var(--icon-card-width);
-  margin-right: calc(-1 * var(--su1));
-  padding: var(--su12) var(--su16);
+  margin-right: var(--size-100-negative);
```
juliodialpad: what's the difference between `size` and `space` variables, here you're using `size` for margin, but in other places you're using `space` so I'm a bit confused?

lib/build/less/utilities/spacing.less
```diff
@@ -49,42 +49,42 @@
...
+.d-stack24 { > * + * { margin-top: var(--space-550) !important; } }
+.d-stack32 { > * + * { margin-top: var(--space-600) !important; } }
+.d-stack48 { > * + * { margin-top: var(--space-650) !important; } }
+.d-stack64 { > * + * { margin-top: var(--space-700) !important; } }
+.d-stack72 { > * + * { margin-top: calc(var(--space-300) * 18) !important; } }
+.d-stack84 { > * + * { margin-top: calc(var(--space-300) * 21) !important; } }
+.d-stack96 { > * + * { margin-top: calc(var(--space-300) * 24) !important; } }
+.d-stack102 { > * + * { margin-top: calc(var(--space-300) * 25.5) !important; } }
+.d-stack114 { > * + * { margin-top: calc(var(--space-300) * 28.5) !important; } }
+.d-stack128 { > * + * { margin-top: calc(var(--space-300) * 32) !important; } }
```
juliodialpad: Why are we doing calculations here again instead of creating the vars for it?

lib/build/less/utilities/spacing.less
```diff
@@ -49,42 +49,42 @@
...
+.d-flow24 { > * + * { margin-left: var(--space-550) !important; } }
+.d-flow32 { > * + * { margin-left: var(--space-600) !important; } }
+.d-flow48 { > * + * { margin-left: var(--space-650) !important; } }
+.d-flow64 { > * + * { margin-left: var(--space-700) !important; } }
+.d-flow72 { > * + * { margin-left: calc(var(--space-300) * 18) !important; } }
+.d-flow84 { > * + * { margin-left: calc(var(--space-300) * 21) !important; } }
+.d-flow96 { > * + * { margin-left: calc(var(--space-300) * 24) !important; } }
+.d-flow102 { > * + * { margin-left: calc(var(--space-300) * 25.5) !important; } }
+.d-flow114 { > * + * { margin-left: calc(var(--space-300) * 28.5) !important; } }
+.d-flow128 { > * + * { margin-left: calc(var(--space-300) * 32) !important; } }
```
juliodialpad: Why are we doing calculations here again instead of creating the vars for it?

juliodialpad (commented):
Are we removing the `spacing units` (@suX) variables? You replaced a lot of them but kept some in other places, what's the difference between `@su`, `--spacing` and `--sizing`, it's not too clear when to use them.

lib/build/less/components/list-group.less
```diff
@@ -17,7 +17,7 @@
 .d-list-group--header,
 .d-list-group--link {
     display: block;
-    padding: var(--size-300) calc(var(--size-300) * 6); // 4 24
+    padding: var(--size-300) var(--size-550); // 4 24
```
braddialpad: padding should be space right?

lib/build/less/utilities/spacing.less
braddialpad: I'm guessing because we want to have only a limited number of space variables available for use in the design system but utility classes need to cover much more than just the core space variables.. however I agree that these calculations are confusing seems like a base value that just gets multiplied by a completely arbitrary number.

lib/build/less/variables/icons.less
```diff
@@ -10,25 +10,25 @@
...
 @icon12:                        1.2rem;
-@icon14:                        1.4rem;
-@icon16:                        1.6rem;
-@icon18:                        1.8rem;
-@icon20:                        2rem;
-@icon24:                        2.4rem;
-@icon32:                        3.2rem;
-@icon48:                        4.8rem;
-@icon64:                        6.4rem;
+@icon14:                        calc(var(--size-500) - var(--size-200));
```
braddialpad, line 26: Again if we're having to do these types of calcs, shouldn't this size just be part of the design system? Something like icon size of 14 feels very core to the design system. If this is because this size is just meant to be specific to icon and that's it, does it make more sense to NOT use size variables and just use rem?

docs/.vuepress/theme/assets/less/dialtone-docs.less
braddialpad: I think this might just be a mistake. margin should be space.

braddialpad (commented):
Just some questions about multiplication calcs similar to julio. Otherwise looks great!

francisrupert:
> Just some questions about multiplication calcs...

Yeah, I'm not a fan of too many magic numbers. I'm reconciling separately in the [Design Tokens first](https://www.figma.com/file/8RbAab7w2rQeg0Ii8jtaAl/DT-Core%3A-Design-Tokens?node-id=3746-13427&t=HOizzAxrNPSJIQiw-11), such that we're supporting legacy values while sticking to classification system of Design Tokens. 

After I get that balanced out in the design tokens, I'll circle back and update here. While this is kind of duplicative work, it will help make the token replacement more straight-forward.

francisrupert:
@braddialpad, @juliodialpad,

Take a look at what I've updated. Rounding out this task is also key to making sure that [DT-1038](https://dialpad.atlassian.net/browse/DT-1038) and [DT-657](https://dialpad.atlassian.net/browse/DT-657) (and probably others) are successful.

Note one thing I haven't done so far in Design Tokens is create them for _all_ current CSS variables we have, just the most important ones. So for the initial series of BETA cuts of `dialtone`'s consumption of Design Tokens will have a mix of everything in `dialtone-tokens` and carry forward those that haven't been mapped out in `dialtone-tokens`.

[DT-1038]: https://dialpad.atlassian.net/browse/DT-1038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DT-657]: https://dialpad.atlassian.net/browse/DT-657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

juliodialpad (approved):
This is much clearer now, a lot of calculations removed and It'd be easier to migrate tokens.
I still need to merge this into #868 and that'd be a lot of fun.

PR #868, opened 2023-04-25
build: convert less mixins to postcss

juliodialpad:
## Description

First clean-up iteration to migrate LESS mixins to PostCSS.

- Removed most of `#internals` usages
- Almost every utility class that was autogenerated with LESS mixins is now on a function within `postcss/dialtone-generators.js`
- `postcss/constants.json` file contains a list of arrays with the needed values, we need to reconcile this with the values on `docs/_data`
- some LESS variables were replaced with CSS variables as PostCSS runs over CSS files not LESS, so I was having issues with that.
- Optimizations to avoid redundancy will be made in the future

## Pull Request Checklist


braddialpad (approved):
Great work on this! I suspect we'll need to make more updates to this as we go.

juliodialpad:
> Great work on this! I suspect we'll need to make more updates to this as we go.

You bet! this is going to be our main entry point to create a lot of stuff from now on.

We could also create all dialtone classes from here if we want, I think that's what `tailwindCSS` is doing currently.

PR #881, opened 2023-05-16
fix(button): set width on label

braddialpad:
## Description
Button width 100% was previously set directly on the Vue component with d-100p. This prevented it from being overridden by utility classes. Moved to Dialtone classes as it should be.

## Pull Request Checklist

### DIFF
--- a/lib/build/less/components/button.less
+++ b/lib/build/less/components/button.less
@@ -188,6 +188,7 @@
     display: inline-flex;
     align-items: center;
     justify-content: center;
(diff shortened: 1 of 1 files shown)


juliodialpad (approved):
Thanks! We should log a ticket to move over all the styles on dialtone-vue that should be on dialtone.

braddialpad:
UPDATE: Changed to flex: 1 auto instead of width. Should have the same result and is more flexboxy.

PR #885, opened 2023-05-26
fix: LESS vars migration

juliodialpad:
## Description

As we removed completely the LESS variables, the script should migrate previous LESS variables to CSS ones.

Changed color migration script that was incorrectly replacing e.g `@black-100` with `@dt-color-black-100` instead of replacing it with `var(--dt-color-black-100)`.

## Pull Request Checklist


docs/.vuepress/theme/components/Home.vue
```diff
@@ -72,210 +73,852 @@
...
-              </g>
+          <rect
+            x="325"
+            y="78"
+            width="322"
+            height="152"
+            rx="4"
+            fill="var(--dt-color-surface-primary)"
+          />
+          <circle
```
braddialpad: hmm I know the linter did this but I kinda hate it... maybe we allow more properties per line?

how about

```
"vue/max-attributes-per-line": ["warn", {
  "singleline": {
    "max": 6
  },      
  "multiline": {
    "max": 1
  }
}]
```

probably revert this file back to its original form before you set this.

braddialpad (changes requested):

braddialpad:
script change itself looks good.

docs/.vuepress/theme/components/Home.vue
juliodialpad: I don't like it either, I'll try those settings

PR #888, opened 2023-06-01
chore: dialtone 8 migration guide

juliodialpad:
## Description

Added migration guide including script usage, recommendations and the list of manual CSS -> Design tokens remapping that are needed to be done after the migration.

**Included some missing colors migrations**

## Pull Request Checklist


migration_guide/Dialtone_8.md
```diff
@@ -0,0 +1,274 @@
+# Dialtone 8 Migration guide
+
+Migration from version `7` to `8`
+
+Dialtone 8.x deprecates the use of LESS variables to favor the usage of Design tokens.
+
+While the migration should be straightforward using the provided script, there might be instances where manual updates
+to a product's code are required.
+These are usually search-and-replace tasks, and may include decisions to be made with your Product Designer; for
+example, RAW HEX values or CSS variables that are not semantically correct.
```
braddialpad: We may want to include an example of a common case where you would need to do this to better illustrate the difference between base / semantic tokens. Perhaps the case I mentioned to you yesterday `--dt-color-neutral-white` to `--dt-color-surface-primary` could work. I see you do this in much more detail down below, but maybe here explain more the concept of a base token and a semantic token.

migration_guide/Dialtone_8.md
```diff
@@ -0,0 +1,274 @@
...
+npm install @dialpad/dialtone@next
+```
+
+### 2. Run the script
+
+- **From the root of your project**
+
+```bash
+npx dialtone-migration-helper --cwd ./path/to/folder
+```
```
braddialpad: I've run into many cases where I have a large folder of components and I only want to migrate a few of the files rather than all of them. In this case I just make a temporary folder called "to_convert", put the files in there, migrate, and then move them back. May want to include how to do this process since I would expect others will have the same question.

braddialpad: We may want to convey a little more that you probably shouldn't run the whole script on the entire codebase all at once, but do one section of the UI, visually verify it and fix any issues and then move on to the next.

Maybe the steps should be like..

Run the script on the desired files for the piece of UI you wish to change
Verify the migration script changed files as expected
Replace css variables with semantic
visually verify the piece of UI you changed in light mode as well as dark mode, fix anything that is obviously off
share preview with a designer and make any further requested changes
make a PR to commit your changes
repeat all this for the next section of UI

migration_guide/Dialtone_8.md
```diff
@@ -0,0 +1,274 @@
...
+|:------------------:|:--------------------------------------------:|:-------------------------------------|
+| `var(--black-500)` |  `var(--dt-color-foreground-placeholder) `   | `var(--dt-color-foreground-disabled` |
+| `var(--black-600)` |    `var(--dt-color-foreground-tertiary)`     |                                      |
+| `var(--black-700)` |    `var(--dt-color-foreground-secondary)`    |                                      |
+| `var(--black-900)` |     `var(--dt-color-foreground-primary)`     |                                      |
+|  `var(--red-300)`  |    `var(--dt-color-foreground-critical)`     |                                      |
+|  `var(--red-400)`  | `var(--dt-color-foreground-critical-strong)` |                                      |
+| `var(--green-400)` |     `var(--dt-color-foreground-success)`     |                                      |
+| `var(--green-500)` | `var(--dt-color-foreground-success-strong)`  |                                      |
+| `var(--gold-500)`  |     `var(--dt-color-foreground-warning)`     |                                      |
```
braddialpad, line 109: May may want to also mention if the actual CSS property is `color` it should be foreground, if the property is `background-color` it should be surface etc. I think you sort of imply that in the heading but we should make that more explicit so there is no confusion.

braddialpad (changes requested):
Lookin great!

migration_guide/Dialtone_8.md
juliodialpad: > I've run into many cases where I have a large folder of components and I only want to migrate a few of the files rather than all of them. In this case I just make a temporary folder called "to_convert", put the files in there, migrate, and then move them back. May want to include how to do this process since I would expect others will have the same question.

I don't find this to be convenient at all, I'd prefer to the add the option to the script to ran into a single file or an array of files, just like `exclude` but inverse.

juliodialpad: > We may want to convey a little more that you probably shouldn't run the whole script on the entire codebase all at once, but do one section of the UI, visually verify it and fix any issues and then move on to the next.
> 
> Maybe the steps should be like..
> 
> Run the script on the desired files for the piece of UI you wish to change Verify the migration script changed files as expected Replace css variables with semantic visually verify the piece of UI you changed in light mode as well as dark mode, fix anything that is obviously off share preview with a designer and make any further requested changes make a PR to commit your changes repeat all this for the next section of UI

I like this steps, I'll add them to the process

braddialpad: > I don't find this to be convenient at all, I'd prefer to the add the option to the script to ran into a single file or an array of files, just like `exclude` but inverse.

Yeah that definitely would be better.

juliodialpad: Adding it

braddialpad (approved):
Fantastic thanks!

PR #914, opened 2023-06-27
feat(datepicker): datepicker doc and style migration

iropolo:
## Datepicker component documentation and style migration

[DLT-1132
](https://dialpad.atlassian.net/jira/software/c/projects/DLT/boards/548?modal=detail&selectedIssue=DLT-1132&assignee=6368be5e3867a77cb8d64d52)


lib/build/less/components/datepicker.less
```diff
@@ -0,0 +1,103 @@
...
+    border: none;
+    border-radius: calc(var(--dt-size-650) + var(--dt-size-200));
+    cursor: pointer;
+
+    &--disabled {
+      color: var(--black-300);
+    }
+
+    &--selected {
+     color: #fff;
```
braddialpad: hmm should this change for dark mode? if so `--dt-color-foreground-primary`

lib/build/less/components/datepicker.less
```diff
@@ -0,0 +1,103 @@
...
+     background: #7c52ff;
```
braddialpad: dt color can be used for this?

lib/build/less/components/datepicker.less
```diff
@@ -0,0 +1,103 @@
...
+
+    > div {
+        display: inline-flex;
+      }
+
+    button {
+      display: inline-flex;
+      align-items: center;
+      padding: 0;
+      color: var(--primary-color);
```
braddialpad: hmm think we shouldn't use --primary-color for anything purple 400 or 500 instead?

lib/build/less/components/datepicker.less
iropolo: replaced it with `var(--dt-color-border-brand)`

lib/build/less/components/datepicker.less
iropolo: replaced it with `var(--purple-400)`

iropolo:
Color migration following guide [Migration Guide Dialtone 8](https://github.com/dialpad/dialtone/blob/version8/migration_guide/Dialtone_8.md)

braddialpad (commented):
Hmm Vue API and example not showing.. Might just need to do the same thing you did on emoji picker.

iropolo:
> Hmm Vue API and example not showing.. Might just need to do the same thing you did on emoji picker.

Yes your right, I didnt push package updates because I did it in emoji-picker PR.
Im guessing if we merge emoji-picker this will work, I dont see it necessary to update it here also.

Lemme merge emoji-picker doc then will update this PR with it.

image

iropolo:
Now It works in the preview 👍🏼

braddialpad (approved):

PR #916, opened 2023-06-28
feat(emoji-picker): emoji-picker doc and style migration

iropolo:
## Emoji-picker component documentation and style migration

[DLT-1090
](https://dialpad.atlassian.net/browse/DLT-1090)

Linked to https://github.com/dialpad/dialtone-vue/pull/1047 update


iropolo:
Fixing lint problems..

lib/build/less/components/emoji-picker.less
```diff
@@ -0,0 +1,224 @@
...
+      padding: calc(var(--dt-size-350) + var(--dt-size-100));
+      background: none;
+      background: rgba(0, 0, 0, 0.1);
+      border: none;
+      border-radius: calc(var(--dt-size-550) + var(--dt-size-300));
+      outline: none;
+      cursor: pointer;
+
+      &:hover {
+        background: #d2d2d2;
```
braddialpad: `var(--dt-color-surface-bold)`

lib/build/less/components/emoji-picker.less
```diff
@@ -0,0 +1,224 @@
...
+
+  &__skin-selected {
+    button {
+      display: inline-block;
+      width: var(--dt-size-625);
+      height: var(--dt-size-625);
+      margin: 0;
+      padding: calc(var(--dt-size-350) + var(--dt-size-100));
+      background: none;
+      background: rgba(0, 0, 0, 0.1);
```
braddialpad: Why is background set twice?

lib/build/less/components/emoji-picker.less
```diff
@@ -0,0 +1,224 @@
...
+    padding: 0 var(--dt-size-500);
+    background: #f9f9f9;
+    border-top: var(--dt-size-100) solid var(--bc-subtle);
+  }
+
+  &__tabset-list {
+      gap: var(--dt-size-300);
+
+      &::after {
+        background-color: var(--black-200) !important;
```
braddialpad: `--dt-color-surface-moderate`

lib/build/less/components/emoji-picker.less
```diff
@@ -0,0 +1,224 @@
...
+        box-shadow: var(--bs-focus-ring);
+      }
+    }
+  }
+
+  &__search {
+    position: relative;
+    z-index: 1;
+    margin: var(--su16) var(--su24) 0 var(--su16);
+    background-color: #fff;
```
braddialpad: `var(--dt-color-surface-primary)`

lib/build/less/components/emoji-picker.less
```diff
@@ -0,0 +1,224 @@
...
+      letter-spacing: -0.01em;
+    }
+  }
+
+  &__category {
+    position: sticky;
+    top: 0;
+    width: 100%;
+    padding-top: var(--dt-size-525);
+    background: rgba(255, 255, 255, 0.9);
```
braddialpad: not going to work in dark mode unless set to semantic color.. Not sure what it should be though.. surface secondary maybe?

lib/build/less/components/emoji-picker.less
iropolo: Seems like a leftover... should linter warn this ? 🤔

lib/build/less/components/emoji-picker.less
iropolo: I didnt did color migrations 🤔 , for size and spacing I used 
https://dialpad.design/tokens/sizing.html
https://dialpad.design/tokens/spacing.html

But if I want to do it with color https://dialpad.design/tokens/color.html for example this `#d2d2d2;` can be 
`var(--dt-color-foreground-secondary)` or
`var(--dt-color-black-700)` but seems like your suggestion `--dt-color-surface-bold` it is different, `#555555`

Before start doing this color migrations, from where I should get this references?, to do it in the right way, cos Im a little bit confused on all this token/styles vars.

Ping @juliodialpad here just in case :)


This question also applies on #914

braddialpad: I just got this via the migration guide https://github.com/dialpad/dialtone/blob/version8/migration_guide/Dialtone_8.md

if you look under Background Color -> Hex values...

Though verify this looks okay.. its a bit strange to me that it maps to a different value. Maybe it's incorrect?

lib/build/less/components/emoji-picker.less
braddialpad: Added this to the linter in my upcoming PR.

lib/build/less/components/emoji-picker.less
braddialpad: I think the hex values here are possibly backwards?

iropolo: I have no idea how to proceed with this colors migrations, I can do it with the migration guide you provide initially and then we can check it later again with the proper migration if it is necessary... 🤔

braddialpad: yeah... 1 sec I'm going to try and verify this

braddialpad: Alright do not use https://dialpad.design/tokens/color this seems to be showing all dark mode tokens currently.. It needs to be updated. Use the migration guide as the source of truth

iropolo: Okay 👍🏼 Will update this PR and #914 with color migration

lib/build/less/components/emoji-picker.less
iropolo: Not the same but close enough, will replace with it `var(--dt-color-surface-secondary)`

iropolo:
Color migration following guide [Migration Guide Dialtone 8](https://github.com/dialpad/dialtone/blob/version8/migration_guide/Dialtone_8.md)

docs/components/emoji-picker.md
```diff
@@ -2,7 +2,27 @@
...
+    :tab-set-labels="tabSetLabels[]"
+    :recently-used-emojis="recentlyUsedEmojis[]"
+    :search-results-label="'Search results'"
+    :search-no-results-label="'Search results'"
+    :search-placeholder-label="'Search...'"
+    @skin-tone="skinTone = $event"
+    @selected-emoji="selectedEmoji"
+  />
+</template>
+```
```
braddialpad, line 93: Are we able to actually render this instead of showing code? (might just need to update dialtone-vue?)

lib/build/less/components/emoji-picker.less
braddialpad: Great, looks good thanks.

docs/components/emoji-picker.md
iropolo: Yes, component render added to the .md file 👍🏼

braddialpad (approved):
Looks great thanks!

PR #926, opened 2023-07-21
fix: make new dialtone util classes higher specificity

braddialpad:
## Description

Because we are running old and new dialtone in ubervoice right now and the utility classes both have the exact same names, we are seeing cases where utility classes are referencing variables from the old dialtone. This change makes the utility classes in the new dialtone higher specificity by doing a double class definition ex: `d-d-flex.d-d-flex` rather than `d-d-flex`. 

We of course don't want to keep this long term and can remove this once DT7 -> DT8 migrations are completed.

## Pull Request Checklist


juliodialpad (changes requested):
These has to do with the import order too on product, right? 

I tested on a clean project and seems to be working properly, but we're missing a bunch of others that would need to have this, every class on the [Utilities folder](https://github.com/dialpad/dialtone/blob/a382785e968690c2ecb05f28d6a3a7b05ffd113b/lib/build/less/utilities) should have this in order to be sure we can use safely every utility class.

braddialpad:
> These has to do with the import order too on product, right?
> 
> I tested on a clean project and seems to be working properly, but we're missing a bunch of others that would need to have this, every class on the [Utilities folder](https://github.com/dialpad/dialtone/blob/a382785e968690c2ecb05f28d6a3a7b05ffd113b/lib/build/less/utilities) should have this in order to be sure we can use safely every utility class.

I made the updates for colors.less. That's the only place we are seeing this issue currently and the only one we really care about at the moment. Will update others if we see more evidence.

juliodialpad (approved):
Looks good, thanks! 

Fix the merge conflicts and don't forget to update the recently added class `.d-bgc-backdrop`.

This is good to go.

PR #938, opened 2023-08-16
fix: bgc color generation

braddialpad:
## Description
Couple issues here, we were not removing the preceding `--` in removePrefixFromColor and we were generating utility classes for all the new theme colors which we don't want, we only want variables.


juliodialpad (approved):
Thanks Brad! I didn't catch that.

PR #940, opened 2023-08-21
fix(input): textarea right space

juliodialpad:
## Description

- Added absolute positioning to right icon on textarea
- Updated documentation examples to `dt-input` component

## Pull Request Checklist

### DIFF
--- a/docs/.vuepress/theme/assets/css/dialtone.css
+++ b/docs/.vuepress/theme/assets/css/dialtone.css
@@ -1,2 +1,0 @@
-/* autoprefixer: off */
-@import "@projectRoot/lib/dist/css/dialtone.css";
--- a/docs/.vuepress/theme/client.js
+++ b/docs/.vuepress/theme/client.js
@@ -3,7 +3,7 @@ import Layout from './layouts/Layout.vue';
 import NotFound from './layouts/NotFound.vue';
 
 // CSS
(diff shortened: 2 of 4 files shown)


iropolo:
hmm, doubt on this but seems like is intended behaviour

image

Would be cool to have a designer review here.

braddialpad:
> hmm, doubt on this but seems like is intended behaviour
> 
> image
> Would be cool to have a designer review here.

Yeah we need to adjust right padding in this case.

iropolo:
> > hmm, doubt on this but seems like is intended behaviour
> > image
> > Would be cool to have a designer review here.
> 
> Yeah we need to adjust right padding in this case.

ye but... that is the actually design...

image

I mean... this PR is for a reason ..

braddialpad:
> ye but... that is the actually design...
> I mean... this PR is for a reason ..

padding will not move the position of the resizer, only changing the width will.

juliodialpad:
- Added padding to the right and increased the right position of the icon to avoid overlapping the scrollbar.
- Loaded dialtone directly from `lib/dist/css/dialtone.css/` again as vuepress wasn't watching the changes, there'd be no problem with the autoprefixer since I added ours to gulp.

braddialpad (approved):
I think this should work reasonably enough for now

PR #941, opened 2023-08-22
fix(item-layout): set width to 100%

braddialpad:
## Description
It was possible for item-layout to overflow it's container. This should lock it to the width of it's container.

## Pull Request Checklist

### DIFF
--- a/lib/build/less/components/item-layout.less
+++ b/lib/build/less/components/item-layout.less
@@ -15,6 +15,7 @@
 .dt-item-layout {
   display: flex;
   align-items: center;
+  width: 100%;
   min-height: calc(var(--dt-size-550) + var(--dt-size-300));
   padding: var(--dt-space-300) var(--dt-space-400);
   font-size: var(--dt-font-size-200);
--- a/lib/build/less/components/item-layout.less
+++ b/lib/build/less/components/item-layout.less
@@ -22,7 +22,7 @@
 
   &--content {
     flex-grow: 1;
-    max-width: 100%;
+    min-width: var(--dt-size-0);
   }
 
   &--subtitle {
--- a/lib/build/less/components/item-layout.less
+++ b/lib/build/less/components/item-layout.less
@@ -15,7 +15,6 @@
 .dt-item-layout {
   display: flex;
   align-items: center;
-  width: 100%;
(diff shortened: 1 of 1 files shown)


iropolo:
I thought maybe `inherit` also could works.. if you want to modified the parent width, this item layout will reacts... but `width: 100%` will force it... lgtm both ways.

iropolo (approved):

juliodialpad:
I'm not so sure about this, let me do some tests before merging.

juliodialpad (changes requested):
`width: 100%` seems to break the story on https://vue.dialpad.design/?path=/story/components-item-layout--default

I'd prefer to use `max-width: 100%`  or `width: inherit` as @iropolo mentioned.

Also, `DtListItem` calculates its size dynamically and `width: 100%` could pontentialy break that functionality.

image

braddialpad:
> I'd prefer to use `max-width: 100%` or `width: inherit` as @iropolo mentioned.

max-width seems to break it in the same way.. not sure if width inherit would fix the initial issue.. I'll try it.

braddialpad:
I will also look at the product code.. maybe they are doing something weird because it appears to take the width of the container when I test it on storybook

iropolo:
for `inherit` it will be necessary to explit set a width value on the parent container. Like `width: 100%`.

braddialpad:
in product they are using d-truncate and that's what's causing this.. don't think it's possible to use truncate without a set width.

braddialpad:
> `width: 100%` seems to break the story on https://vue.dialpad.design/?path=/story/components-item-layout--default

Okay I set max-width 100% on the inner content element rather than the whole thing. This seems to fix the truncation problem and doesn't seem to break anything existing.
Screenshot 2023-08-22 at 2 44 54 PM

braddialpad:
Nevermind... still not fixed. Now it doesn't highlight correctly.

Screenshot 2023-08-22 at 2 56 41 PM

braddialpad:
Alright got it. `min-width: 0` was the secret.

from this article: https://css-tricks.com/flexbox-truncated-text/

Screenshot 2023-08-22 at 3 22 39 PM

PR #951, opened 2023-09-18
feat(migration-script): support var prefixed with -

braddialpad:
## Description
Couple things done here to the migration script:

- Added in the case for variables like `-@su4` to be replaced with `var(--dt-space-300-negative)` as these are fairly common in product
- Changed `@su` and `--su` variables to be replaced with space instead of size. This still does not guarantee correctness as su variables could still be used for size. However su variables are used much more commonly for space rather than size, so there should be less manual fixes that have to be done.

## Pull Request Checklist


juliodialpad (approved):
Great work here!

I'd prefer to have 1 regex to avoid repetitive code, here's a regex I was testing and might work.

`/-@su1\b|-var\(--su1\)|[^-]var\(--sun1\)/g`

But if you've tested and the way you made it works better I'm ok with that. Thanks!

braddialpad:
> Great work here!
> 
> I'd prefer to have 1 regex to avoid repetitive code, here's a regex I was testing and might work.
> 
> `/-@su1\b|-var\(--su1\)|[^-]var\(--sun1\)/g`
> 
> But if you've tested and the way you made it works better I'm ok with that. Thanks!

Nice, yeah that's better. Works for all the tests I tried. Replaced it.

PR #957, opened 2023-09-26
docs: extract constants

juliodialpad:
## Description

Extract and provide dialtone-vue constants so they're available to use on doc site.

## Pull Request Checklist

### DIFF
--- a/docs/.vuepress/theme/client.js
+++ b/docs/.vuepress/theme/client.js
@@ -47,11 +47,27 @@ export default defineClientConfig({
 
 async function registerDialtoneVue (app) {
   const module = await import('@dialpad/dialtone-vue');
-  const dialtoneComponents = Object.keys(module).filter((key) => key.startsWith('Dt'));
-  dialtoneComponents.forEach((key) => {
-    app.component(key, module[key]);
+  const {
+    dialtoneComponents,
+    dialtoneConstants,
+  } = Object
+    .keys(module)
+    .reduce(
+      (result, key) => {
+        if (key.startsWith('Dt')) {
+          result.dialtoneComponents.push({ [key]: module[key] });
+        } else if (/^[A-Z_]+$/.test(key)) {
+          result.dialtoneConstants.push({ [key]: module[key] });
+        }
+        return result;
+      },
+      { dialtoneComponents: [], dialtoneConstants: [] },
+    );
+  Object.keys(dialtoneComponents).forEach(key => {
+    app.component(key, dialtoneComponents[key]);
   });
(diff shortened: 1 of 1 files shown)


docs/.vuepress/theme/client.js
```diff
@@ -48,10 +48,18 @@ export default defineClientConfig({
 async function registerDialtoneVue (app) {
   const module = await import('@dialpad/dialtone-vue');
   const dialtoneComponents = Object.keys(module).filter((key) => key.startsWith('Dt'));
-  dialtoneComponents.forEach((key) => {
+  const dialtoneConstants = Object
+    .keys(module)
+    .filter((key) => /^[A-Z_]+$/.test(key))
```
braddialpad, line 53: Wouldn't this also pull in components starting with DT?

docs/.vuepress/theme/client.js
```diff
@@ -48,10 +48,18 @@ export default defineClientConfig({
...
+    .reduce((res, key) => {
+      res[key] = module[key];
+      return res;
+    }, {});
+  dialtoneComponents.forEach(key => {
     app.component(key, module[key]);
   });
   app.provide('dialtoneComponents', dialtoneComponents);
+  app.provide('dialtoneConstants', dialtoneConstants);
```
braddialpad: Doesn't this only provide to the template? Is it possible to use this in only javascript?

braddialpad (commented):
Just some questions.

docs/.vuepress/theme/client.js
juliodialpad: In case there's a component that starts with DT, yes. But all the components should start with Dt. This filters only strings that start and end with capital letters or _

docs/.vuepress/theme/client.js
juliodialpad: What do you mean by only javascript? 🤔

braddialpad: Can you just use the constant within `<script>` tags without an import?

docs/.vuepress/theme/client.js
braddialpad: got ya. should be alright then.

docs/.vuepress/theme/client.js
juliodialpad: > Can you just use the constant within `<script>` tags without an import?

I don't think I'm understanding how can I achieve that

juliodialpad: You mean something like this?

```js
window.dialtoneConstants = dialtoneConstants;
```

braddialpad (approved):

PR #961, opened 2023-10-02
feat(stack): add defaults for align prop

ninamarina:
## Description
https://dialpad.atlassian.net/browse/DLT-1265
Adds defaults for aligning items in the Stack component.
Documentation will be updated in the `dialtone-vue` repository.

## Pull Request Checklist

### DIFF
--- a/lib/build/less/components/stack.less
+++ b/lib/build/less/components/stack.less
@@ -18,14 +18,12 @@
   &--column {
     --stack-direction: column;
 
-    align-items: normal;
     justify-content: flex-start;
   }
 
   &--column-reverse {
     --stack-direction: column-reverse;
 
-    align-items: normal;
     justify-content: flex-start;
   }
 
@@ -51,7 +49,6 @@
   display: flex;
   flex-direction: var(--stack-direction);
   gap: var(--stack-gap);
-  align-items: normal;
   justify-content: flex-start;
 }
 
@@ -63,7 +60,6 @@
 .d-stack--column-reverse {
   --stack-direction: column-reverse;
 
-  align-items: normal;
   justify-content: flex-start;
 }
 
--- a/lib/build/less/components/stack.less
+++ b/lib/build/less/components/stack.less
@@ -17,34 +17,42 @@
 .direction-options() {
   &--column {
     --stack-direction: column;
-    --align-items: flex-start;
+
+    align-items: normal;
+    justify-content: flex-start;
   }
 
   &--column-reverse {
     --stack-direction: column-reverse;
-    --align-items: flex-start;
+
+    align-items: normal;
+    justify-content: flex-start;
   }
 
   &--row {
     --stack-direction: row;
-    --align-items: center;
+
+    align-items: center;
+    justify-content: normal;
   }
 
   &--row-reverse {
     --stack-direction: row-reverse;
-    --align-items: center;
+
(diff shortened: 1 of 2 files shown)


juliodialpad (changes requested):
I think we don't need to have those as utility classes.

I'd be better to just set those default values for row and column variants of the stack component.

And just say in the documentation that by default row stack will have `align-items: center` and column stack has `justify-content: flex-start` set by default.

ninamarina:
> I think we don't need to have those as utility classes.
> 
> I'd be better to just set those default values for row and column variants of the stack component.
> 
> And just say in the documentation that by default row stack will have `align-items: center` and column stack has `justify-content: flex-start` set by default.

Changed in 3353c31b30dff7b8467815fd54323fee3038d21a

lib/build/less/components/stack.less
```diff
@@ -17,18 +17,30 @@
 .direction-options() {
   &--column {
     --stack-direction: column;
+
+    align-items: normal;
```
iropolo: `normal` is the default value of `align-items` Is this adding value here?

lib/build/less/components/stack.less
```diff
@@ -17,18 +17,30 @@
...
+    justify-content: flex-start;
   }
 
   &--row {
     --stack-direction: row;
+
+    align-items: center;
+    justify-content: normal;
```
iropolo, line 34: is this `justify-content: normal;` css property doing some effect?

iropolo (approved):
pre-approving it.

Have some doubt on the use of `align-items: normal` and `justify-content: normal` but if it is intended to solve some issue or edge case would be cool to review it here, if it is not I would remove them.

Thanks,

lib/build/less/components/stack.less
ninamarina: It's to override the default defined here https://github.com/dialpad/dialtone/pull/961/files#diff-a6abaef1f0bdd89f14149d49806e8d9d66c282871015d46657d8ff6b988a8566R55

lib/build/less/components/stack.less
ninamarina: Maybe for the `align-items` I don't need it, but I do for `justify-content`, so I thought for clarity and consistency add both

lib/build/less/components/stack.less
iropolo: duh, right, didnt see it is under `&--row {` 👍🏼

lib/build/less/components/stack.less
iropolo: Would remove it if it is not adding value. 
in the other hand, `justify-content` is adding something necessary to work correctly.
Up to you, looks good to me I already approved it 👍🏼

ninamarina: I take your suggestion, let's have the less lines of css we can 👍

braddialpad (approved):
Looks good!

juliodialpad (approved):
Looking good, thanks Nina!

I'll create a ticket to improve this component, instead of using media queries for responsive versions we should be using our dialtone generators function but that's another story. I'll create the ticket and refer back to this PR.

juliodialpad:
Cleanup stack component ticket: https://dialpad.atlassian.net/browse/DLT-1270

PR #967, opened 2023-10-06
feat(typography): add utility for word wrap

fabiangarga:
## Description
Add utils for word wrap

## Pull Request Checklist

### DIFF
--- a/docs/utilities/typography/word-wrap.md
+++ b/docs/utilities/typography/word-wrap.md
@@ -1,6 +1,6 @@
 ---
 title: Word wrap
-description: Utilities for controlling the way words wrap within an element.
+description: Utilities for controlling the way words wrap within an element. Generally used for handling overflow of long strings that are actually supposed to be a single unbroken word, like URLs or file paths.
 ---
 
 ## Normal
@@ -19,7 +19,7 @@ Use `d-ww-normal` to break words only at allowed break points.
 
 ## Break word
 
-Use `d-ww-break-word` to allow unbreakable words to be broken.
+Use `d-ww-break-word` to allow unbreakable words to be broken. Is a more conservative approach than [`d.wb-break-all`](./word-break.md#break-all) and will only break long words that do not fit the container.
 
 <code-well-header class="d-fl-center d-p24 d-bgc-green-100 d-bgo50 d-w100p d-hmn102" custom>
   <div class="d-bgc-green-100 d-py8 d-px16 d-bar8 lg:d-w216 d-w332">
--- a/docs/utilities/typography/word-wrap.md
+++ b/docs/utilities/typography/word-wrap.md
@@ -1,0 +1,73 @@
+---
+title: Word wrap
+description: Utilities for controlling the way words wrap within an element.
+---
+
+## Normal
+
+Use `d-ww-normal` to break words only at allowed break points.
+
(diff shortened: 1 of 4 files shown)


iropolo (approved):
Pre-approved.
Looks goods, thanks.
Would be cool to have a design review here.

docs/utilities/typography/word-wrap.md
```diff
@@ -0,0 +1,73 @@
+---
+title: Word wrap
```
francisrupert, line 2: Between **Word wrap** and **Word break** it would be good to call out the difference between them. Here's how I've described it before...

* `word-break: break-all;` will break _between_ any two characters, which can lead to awkward line breaks in the middle of short words. 
* `word-wrap: break-word;` is more conservative and will only break long words that do not fit the container.
* `word-wrap` is generally better for handling overflow of long strings that are actually supposed to be a single unbroken word, like URLs or file paths.
* `word-break` is more aggressive and is useful when you want to ensure that text will absolutely not overflow its container, regardless readability risk.

Take into consideration the above and integrate between those two Utility doc pages, including cross-linking.

braddialpad (approved):
Looks good thanks. Good to go after making Francis' changes.

docs/utilities/typography/word-wrap.md
fabiangarga: Sounds good, just updated the doc with this info and added cross-references to `word-wrap` <=> `word-break`

Demo of docs:
https://github.com/dialpad/dialtone/assets/5950950/be860a4b-cd66-4ca3-9ec4-63a051a873e2

lib/build/less/utilities/typography.less
```diff
@@ -273,6 +273,13 @@ ul {
 .d-wb-keep-all { word-break: keep-all !important; }
 .d-wb-unset { word-break: unset !important; }
 
+//  ============================================================================
+//  $$  WORD WRAP
+//  ----------------------------------------------------------------------------
+.d-ww-break-word { word-wrap: break-word; }
```
ninamarina: I see the `!important` was not added here, as it is in all other utilities, and also the docs show it. I don't know if we should add it here, or remove it from everywhere else, since I am not sure what is it's purpose and I am generally biased towards not using it, what do you think @braddialpad ?

braddialpad: Yes all utility classes should be important because they are immutable https://csswizardry.com/2016/05/the-importance-of-important/

But you're right we don't want to use important for anything else.

fabiangarga: Updated with this, thanks both 🙏 

@ninamarina I agree, since it forces the immutable approach to be implemented as well in any of the properties that we might want to overrite with a higher specificity value. But its definitely useful to have the utils as immutables, not sure if all the utils, or maybe when using the utilities we should be able to specify the Immutability?

fabiangarga:
Updated this branch with the changes requested, let me know if it is ok to merge and will proceed. Thanks @braddialpad @ninamarina @iropolo @francisrupert  🙏

braddialpad (approved):

braddialpad:
Good to go, thanks.

francisrupert:
Thanks @fabiangarga!

PR #973, opened 2023-10-12
docs: update images in components overview for light and dark support

yorbi-dp:
## Description
Updated all images in the Components Overview to SVG format to ensure compatibility with both light and dark modes.

## Pull Request Checklist

overview_components


francisrupert:
Definitely looks like you'll need guidance on the best Vue method for inserting the `<svg>` directly inline. The PR deploy preview shows 404 error for each one, which is probably a simple fix in and of itself for proper path. However, inspecting reveals it is prefetching every single image asset or something and inserting a whole lot of redundant stuff and I'm unclear of the performance impact. It did nearly crash my system just trying to inspect.

image

https://github.com/dialpad/dialtone/assets/1165933/4a28f790-1d0d-4443-92d4-ff82662464bc

yorbi-dp:
Oh oh, just saw that in the preview hah. Already checking on it.
Looks like it's fetching every single .svg in the /assets/images/ folder for every card.
Going to take a double look on that. 
Thank you!

francisrupert:
> ...prefetching every single image asset or something and inserting a whole lot of redundant stuff...

I can walk back this comment a bit. I pulled it down locally and this doesn't occur. The path issue probably makes sense given the inclusion of `/deploy-previews/pr-973/`.

Meanwhile, I defer to others on the Vue technique at play. 

But dang, this looks really good in light and dark. I'll review the actual artwork separately.

francisrupert:
There's a bit of an empty space flash before they're injected in. From a Vue perspective I wonder if there's a way for them to render sooner, or at the vert least hold that space before they're injected — keeping in mind it's not a fluid size.

https://github.com/dialpad/dialtone/assets/1165933/00a9fbe2-7724-43be-995e-b621473fc9c2

yorbi-dp:
Instead of fetching the SVG images from the intended PR preview folder at https://dialpad.design/deploy-previews/pr-973/components/, is attempting to load them from the production folder https://dialpad.design/assets/images/components/. This misdirection is leading to a 404 error. The response is being rendered within the small image container as a "Not Found 404" page. This likely accounts for the redundancy and performance issues we've been observing in the preview.

For reference, I'm using an absolute path to fetch the images: 
`svg: /assets/images/${props.basePath}/${page.fileName}.svg`.

Now I'm adding the PNG as a fallback image when there is no SVG, in case this happens in the future.

> From a Vue perspective I wonder if there's a way for them to render sooner, or at the vert least hold that space before they're injected — keeping in mind it's not a fluid size. - @francisrupert 

To hold the space of the image if it's not render yet, we can add a min-height to the wall-image class .

Will keep the thread updated with any findings or further assistance needed!

francisrupert:
> ...To hold the space of the image if it's not render yet, we can add a min-height to the wall-image class...

Keep in mind it's a fluid width and height, so consider how to do so in a _proportional_ way. 

https://github.com/dialpad/dialtone/assets/1165933/675e6baf-3b4e-4ffa-9471-b45b51cadf8c

docs/.vuepress/views/Overview.vue
```diff
@@ -13,11 +13,8 @@
           v-if="page.thumb"
           class="dialtone-wall__image"
         >
-          <img
-            class="dialtone-wall__thumb"
-            :alt="`${page.fileName}-thumbnail`"
-            :src="$withBase(`/assets/images/${basePath}/${page.fileName}.png`)"
-          >
+          <!--<div v-if="backgroundImage" class="dialtone-wall__background" v-html="backgroundImage"></div>-->
+          <div v-if="svgImage[index]" class="dialtone-wall__thumb" v-html="svgImage[index]"></div>
```
braddialpad: Couple things

Not relevant anymore now that I found we need inline svgs.

~~1. We should try not to use v-html here, you should try and use v-html as little as possible. I'm not sure we need to worry about asynchronous fetching here at all either, seems like we're trying to do far too much for something very simple. svg's can be rendered using the `img` tag. You can probably get rid of all the fetching code below and use a structure similar to before.~~

```
<img class="dialtone-wall__thumb" :alt="`${page.fileName}-thumbnail`" :src="`/assets/images/${props.basePath}/${page.fileName}.svg`">
```

2. Whenever we reference public assets we need to use the `$withBase` function, this is what handles the path for the dynamic base address ex/ `dialpad.design/deploy-previews/pr-722` and is what is causing that 404 error you're getting.

```
<img class="dialtone-wall__thumb" :alt="`${page.fileName}-thumbnail`" :src="$withBase(`/assets/images/${props.basePath}/${page.fileName}.svg`)">
```

braddialpad (changes requested):

braddialpad:
There's a bunch of linter errors, I suggest you turn on autofix lint errors on save in your IDE. Looks like we don't autofix on commit on dialtone, only on dialtone-vue. I'll make a PR to change this.

docs/.vuepress/views/Overview.vue
francisrupert: @braddialpad 

> ...rendered using the img tag...

SVGs via `` don't support CSS variables. Yorbi's intention is to inject the `<svg>` inline. What would be the best way to do that? Perhaps a Vue component (say, `dialtone/docs/.vuepress/theme/components/InlineSVG`). Kind of like the Vue icon component. Whatever is simplest and scalable. 

This might be the presumptive method throughout the doc site going forward; that is, we anticipate displaying more visuals throughout and they may often be `<svg>`.

braddialpad: Ah I see, thought I must have been missing something here.

I will explore this, give me a min

braddialpad: Alright we should be able to use this nice little plugin that allows you to render svgs as vue components https://www.npmjs.com/package/vite-svg-loader 

I tested it out works really nice. I'll add it to our package, then you can merge in my change and use it.

to render an svg inline simply use like so:

```
<template>
  <IconComponent />
</template>

import IconComponent from './my-icon.svg?component'
```

because you are rendering a list it'll be a bit harder than that. You'll probably want to dynamically import all the svgs from a folder:

```
const svgs = [];

beforeMount () {
  // get all file paths from folder here

  // loop over the below statement to dynamically import all files
  import(`${filePath}/${fileName}.svg?component`).then((svg) => {
    this.svgs.push(svg);
  });
},
```

In the template you'll have to render using a dynamic component and find some way to map the page to a component, perhaps by making the SVG component names exactly the same as the page title, or creating a mapping object if that's not possible.

```
<template
  v-for="(page, index) in pages"
  :key="page.title"
>
  // page.title likely won't work without some kind of manipulation
  <component :is="page.title">
</template>
```

braddialpad: @yorbi-dp Spoke with francis about this a bit and I've added a component to render an svg inline in this PR. https://github.com/dialpad/dialtone/pull/977

yorbi-dp: > @yorbi-dp Spoke with francis about this a bit and I've added a component to render an svg inline in this PR. #977

Awesome! This will be helpful for using svg images across the docs site (and probably Dialpad as well), making them compatible with light/dark modes without any additional adjustments.

Going to take a look and update the PR with this new component.

Thanks! 🙌

yorbi-dp:
> There's a bunch of linter errors, I suggest you turn on autofix lint errors on save in your IDE. Looks like we don't autofix on commit on dialtone, only on dialtone-vue. I'll make a PR to change this.

Thank you!
I also missed running `fix:lint` before pushing.
I'll make sure to turn "autofix lint errors" for my next push.

docs/.vuepress/views/Overview.vue
```diff
@@ -1,23 +1,18 @@
...
         class="dialtone-wall__item"
       >
-        <div
-          v-if="page.thumb"
-          class="dialtone-wall__image"
-        >
-          <img
+        <div v-if="page.thumb" class="dialtone-wall__image">
+          <svg-loader class="dialtone-wall__thumb" :name="page.fileName" />
+          <!-- <img
```
ninamarina: You can delete this commented tag

ninamarina (approved):
Amazing work! 😍

docs/.vuepress/public/assets/images/components/banner.svg
```diff
@@ -0,0 +1,23 @@
+<svg width="264" height="107" viewBox="0 0 264 107" fill="none" xmlns="http://www.w3.org/2000/svg">
```
francisrupert: Each `/assets/images/components/file.svg` should be fluid. Currently they have an inconsistent mix of `width`, `height` and `viewBox` attributes.

`width` and `height` are unnecessary as long as the `viewBox` value is identical between all. That's probably `viewBox="0 0 264 107"` ... (Unsure why I did `107` w/ previous `png` thumbs, such a random number).

francisrupert (changes requested):

docs/.vuepress/public/assets/images/components/banner.svg
yorbi-dp: Ooh that makes sense! I've included the 100% width in most of them, but I'll certainly remove it since it's already defined in the viewBox. Thanks for pointing that out! Going to create a fix for that.

docs/.vuepress/views/Overview.vue
yorbi-dp: Thanks Nina, just remove it!
I'll work on a future version with a fallback img in case the SVG it's not found or couldn't render.

docs/.vuepress/public/assets/images/components/banner.svg
yorbi-dp: Fixed in the lastest commit 🙌

braddialpad (approved):
Looks good by me, thanks!

docs/.vuepress/public/assets/images/components/avatar.svg
```diff
@@ -1,5 +1,5 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="100%" height="107" viewBox="0 0 264 107" fill="none" id="avatar-shape">
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 264 107" fill="none" id="avatar-shape">
```
francisrupert: `id="avatar-shape"`

Can probably remove all the `id` attributes on the `<svg>`, right? Maybe a consequence of your WIP plugin? 

Other than that, I approve; ship it!

francisrupert (approved):

docs/.vuepress/public/assets/images/components/avatar.svg
yorbi-dp: Great! Yes, it was added by the plugin lol. Fixed! 

Also fixed the plugin so it doesn't add the `id` and removes the `width` and `height` attributes from the <svg> tag. 

Thanks!

PR #982, opened 2023-10-20
docs: add tokens examples

ninamarina:
## Description
Jira ticket: https://dialpad.atlassian.net/browse/DLT-1186

Adds visual examples for the tokens page. Affects Color, Typography, Shadow, Sizing, Spacing.

## Pull Request Checklist

tokens-examples

### DIFF
--- a/docs/.vuepress/baseComponents/TokenExample.vue
+++ b/docs/.vuepress/baseComponents/TokenExample.vue
@@ -1,0 +1,103 @@
+<template>
+  <div
+    v-if="category === 'color'"
+    class="d-bar-circle d-w42 d-h42"
+    :style="getColorStyle()"
+  />
+  <p
+    v-if="category === 'typography'"
+    :style="getTypographyStyle()"
+  >
+    Aa
+  </p>
+  <div
+    v-if="category === 'shadow'"
+    class="d-bar2 d-w42 d-h42"
+    :style="getShadowStyle()"
+  />
+</template>
+
+<script>
+import { CATEGORY_MAP } from './TokenTable.vue';
+
+const TYPOGRAPHY_KEY_MAP = {
+  family: 'font-family',
+  size: 'font-size',
+  'line-height': 'line-height',
+  weight: 'font-weight',
+  'text-case': 'text-transform',
+};
+
+const SHADOW_COMPOSITION_TOKENS = ['small', 'medium', 'large', 'extra-large', 'card', 'focus', 'focus-inset'];
+
+export default {
+  name: 'TokenExample',
+
+  props: {
+    category: {
+      type: String,
+      default: 'color',
+      validator: (v) => Object.keys(CATEGORY_MAP).includes(v),
+    },
+
+    name: {
+      type: String,
+      default: '',
+    },
+
+    value: {
+      type: String,
+      default: '',
+    },
+  },
+
+  methods: {
+    getColorStyle () {
+      const property = /^var\(--dt-(\w+)-/.exec(this.name)[1];
+      switch (property) {
+        case 'color':
+          return `background: ${this.value}`;
+
+        case 'opacity':
+          return `background: rgba(0, 0, 0, ${this.value})`;
+
+        case 'theme':
+          return `background: ${this.value}`;
+
+        default:
+          return null;
+      }
+    },
+
+    getTypographyStyle () {
+      const fontRegex = new RegExp(`^var\\(--dt-font-(${Object.keys(TYPOGRAPHY_KEY_MAP).join('|')})`);
+      const fontProperties = fontRegex.exec(this.name);
+      if (fontProperties) {
+        return `${TYPOGRAPHY_KEY_MAP[fontProperties[1]]}: ${this.value}`;
+      }
+      const typographyRegex = new RegExp(`^var\\(--dt-typography-.+(${Object.keys(TYPOGRAPHY_KEY_MAP).join('|')})\\)$`);
+      const typographyProperties = typographyRegex.exec(this.name);
+      if (typographyProperties) {
+        return `${TYPOGRAPHY_KEY_MAP[typographyProperties[1]]}: ${this.value}`;
+      }
+      const isCompositionToken = this.name.startsWith('var(--dt-typography');
+      if (isCompositionToken) {
+        return `font: ${this.value}`;
+      }
+      const compositionRegex =
+        new RegExp(`^var\\(--dt-typography-.+(${Object.keys(TYPOGRAPHY_KEY_MAP).join('|')})\\)$`);
+      if (compositionRegex) {
+        return `font: ${this.value}`;
+      }
+      return null;
+    },
+
+    getShadowStyle () {
+      if (SHADOW_COMPOSITION_TOKENS.find(name => this.name.endsWith(`${name})`))) {
+        return `box-shadow: ${this.value}`;
+      }
+      return null;
+    },
+  },
+};
+</script>
--- a/docs/.vuepress/baseComponents/TokenExample.vue
+++ b/docs/.vuepress/baseComponents/TokenExample.vue
@@ -21,14 +21,18 @@
     :style="getSizeStyle"
   />
   <div v-if="category === 'space'" class="space">
-    <div v-if="displaySpaceReference" class="spaceReference" :style="getSpaceStyle">
+    <div v-if="displaySpaceReference" :class="[{ percentage: isPercentage }, 'spaceReference', 'spaceBefore']">
       A
     </div>
     <div
       class="rectangle"
       :style="getSizeStyle"
     />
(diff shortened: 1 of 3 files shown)


docs/.vuepress/baseComponents/TokenExample.vue
```diff
@@ -0,0 +1,123 @@
...
+      if (SHADOW_COMPOSITION_TOKENS.find(name => this.name.endsWith(`${name})`))) {
+        return `box-shadow: ${this.value}`;
+      }
+      return null;
+    },
+
+    getSizeStyle () {
+      if (this.value.endsWith('%')) return null;
+      const value = parseFloat(this.value.replace('rem', ''));
+      if (value < 12.8 && value > 0) return `width: ${this.value}`;
```
iropolo: `if (value < 12.8 && value > 0) return { width: this.value };`

docs/.vuepress/baseComponents/TokenExample.vue
```diff
@@ -0,0 +1,123 @@
...
+      default: '',
+    },
+
+    value: {
+      type: String,
+      default: '',
+    },
+  },
+
+  methods: {
```
iropolo: would update it to `computed` to ensure reactivity.

docs/.vuepress/baseComponents/TokenExample.vue
```diff
@@ -0,0 +1,123 @@
...
+      }
+      const isCompositionToken = this.name.startsWith('var(--dt-typography');
+      if (isCompositionToken) {
+        return `font: ${this.value}`;
+      }
+      return null;
+    },
+
+    getShadowStyle () {
+      if (SHADOW_COMPOSITION_TOKENS.find(name => this.name.endsWith(`${name})`))) {
```
iropolo: Will update `find` method to `some` method, I see more logical in the process since it only matters if there is a match or not... we are not searching nothing in particular.

docs/.vuepress/baseComponents/TokenExample.vue
```diff
@@ -0,0 +1,123 @@
...
+        return `box-shadow: ${this.value}`;
```
iropolo: `return { 'box-shadow': this.value };`

docs/.vuepress/baseComponents/TokenExample.vue
```diff
@@ -0,0 +1,123 @@
...
+      const typographyProperties = typographyRegex.exec(this.name);
+      if (typographyProperties) {
+        return `${TYPOGRAPHY_KEY_MAP[typographyProperties[1]]}: ${this.value}`;
+      }
+      const isCompositionToken = this.name.startsWith('var(--dt-typography');
+      if (isCompositionToken) {
+        return `font: ${this.value}`;
+      }
+      return null;
+    },
```
iropolo:
```
getTypographyStyle() {
      for (const key in TYPOGRAPHY_KEY_MAP) {
        if (this.name.includes(`--dt-font-${key}`) || this.name.includes(`--dt-typography-${key}`)) {
          return { [TYPOGRAPHY_KEY_MAP[key]]: this.value };
        }
      }
      if (this.name.startsWith('var(--dt-typography')) {
        return { font: this.value };
      }
      return null;
    },
```

docs/.vuepress/baseComponents/TokenExample.vue
```diff
@@ -0,0 +1,123 @@
...
+        case 'opacity':
+          return `background: rgba(0, 0, 0, ${this.value})`;
+
+        case 'theme':
+          return `background: ${this.value}`;
+
+        default:
+          return null;
+      }
+    },
```
iropolo: Will update `property` to `this.name.split('--dt-')[1]?.split('-')[0];`

Will unify 'color' and 'theme' since it is the same return

```
 case 'color':
        case 'theme':
          return { background: this.value };
```

Will update 'opacity' return to 
`return { background: `rgba(0, 0, 0, ${this.value})` };
`

iropolo (approved):
I see it good and working well. Approved from my side, would be cool to have designer review.

My suggestions are just nitpick.
Curious about why we are using <script> in some components and <script setup> in some others, probably because its just old, would suggest to unify this in the future, I will bring this to discuss.

thanks!!!

braddialpad:
> probably because its just old

Yes that is exactly the reason, we should maybe convert all existing documentation components to script setup since they are all vue 3

docs/.vuepress/baseComponents/TokenExample.vue
ninamarina: fixed in 2f021742c8882a75815c7d36af27fd8669108c4d

docs/.vuepress/baseComponents/TokenExample.vue
ninamarina: updated in 2f021742c8882a75815c7d36af27fd8669108c4d

docs/.vuepress/baseComponents/TokenExample.vue
ninamarina: agree! changed in 2f021742c8882a75815c7d36af27fd8669108c4d

docs/.vuepress/baseComponents/TokenExample.vue
ninamarina: fixed in 2f021742c8882a75815c7d36af27fd8669108c4d

docs/.vuepress/baseComponents/TokenExample.vue
ninamarina: this is much simpler than the regex I used, thank you! had to tweak this code a little, but changed in 2f021742c8882a75815c7d36af27fd8669108c4d

docs/.vuepress/baseComponents/TokenExample.vue
ninamarina: fixed in 2f021742c8882a75815c7d36af27fd8669108c4d

docs/.vuepress/baseComponents/TokenExample.vue
```diff
@@ -0,0 +1,123 @@
...
+  <div
+    v-if="category === 'shadow'"
+    class="d-bar2 d-w42 d-h42"
+    :style="getShadowStyle()"
+  />
+  <div
+    v-if="category === 'size'"
+    class="rectangle"
+    :style="getSizeStyle()"
+  />
```
braddialpad: For radius tokens we should probably show a square or rectangle with rounded edges like so: https://polaris.shopify.com/tokens/border

francisrupert:
Added @yorbi-dp as a design reviewer, especially as we know he and Fede will have grander vision for how this is all represented in other tasks.

docs/.vuepress/baseComponents/TokenExample.vue
ninamarina: Absolutely, and for borders I should probably show the border of the rectangle, or a line like in that page. @yorbi-dp what do you think?

ninamarina:
> > probably because its just old
> 
> Yes that is exactly the reason, we should maybe convert all existing documentation components to script setup since they are all vue 3

Changed it to script setup in 4f5c6c32776ed48fe75e173e78c156dc5d0a7914

braddialpad (approved):
I'm good on this, there is still the issue of the border radius but we can fix later when we get feedback from design if necessary.

docs/.vuepress/baseComponents/TokenExample.vue
yorbi-dp: Agree! Couple things we can improve.

- For borderRadius: We can use a fixed `var(--dt-size-625)` x `var(--dt-size-625)` square with the different borderRadius.
image

- For the stroke: we can same squares and add the `var(--dt-color-border-brand)` as the color and with no surface/background color.

- To demonstrate the use of space: we can include two elements before and after the shape to represent the spacing, similar to this example:
image

- Why are sizes from `var(--dt-size-800)` to `var(--dt-size-1200)` not rendering as a full rectangle?
- We shouldn’t display the var(--dt-size-base) and var(--dt-space-base) and var(--dt-font-size-root) variables since it’s for an internal use only.

Please feel free to reach out if you have any questions @ninamarina!

ninamarina: Thank you @yorbi-dp, I will reach you with a couple of questions, but to answer 

> Why are sizes from var(--dt-size-800) to var(--dt-size-1200) not rendering as a full rectangle?

this is because they were extremely big.

ninamarina: @yorbi-dp I updated the PR with these suggestions if you want to check it out

yorbi-dp:
Thank you Nina! This looks great, as we look toward the future of the `tokens` section!

yorbi-dp (approved):
Thank you Nina! This looks great, as we look toward the future of the tokens section!

PR #985, opened 2023-10-20
fix: remove z-index on root-layout

braddialpad:
## Description
Issue on the doc site with hero images overlapping the topbar. Removed the z-index from these images, since it really served no purpose in the first place, and then also removed z-index from the sticky root layout class. It is best to avoid use of z-index if necessary, and here it does not seem necessary. The natural dom flow should work fine in the root layout case. If a product deems it necessary to have z-index on nav elements, they are able to add it at the product level via class props.

## Pull Request Checklist


juliodialpad (approved):
Good to get rid of unnecessary z-index, thanks!

